PR #1, opened 2022-05-25
backport the change of Claudio to LTS version

oliviermattelaer:
So this is a proposal received from Claudio to have better control on the number of thread used by the code.

### DIFF
--- a/madgraph/interface/amcatnlo_run_interface.py
+++ b/madgraph/interface/amcatnlo_run_interface.py
@@ -1870,8 +1870,6 @@ Please read http://amcatnlo.cern.ch/FxFx_merging.htm for more details.""")
             # We are done.
             self.finalise_run_FO(folder_names[mode],jobs_to_collect)
             self.update_status('Run complete', level='parton', update_results=True)
-            if hasattr(self, 'cluster') and isinstance(self.cluster, cluster.MultiCore):
-                self.cluster.remove()
             return
 
         elif mode in ['aMC@NLO','aMC@LO','noshower','noshowerLO']:
@@ -2985,10 +2983,12 @@ RESTART = %(mint_mode)s
                 logger.warning('Impossible to detect the number of cores => Using One.\n'+
                         'Use set nb_core X in order to set this number and be able to'+
                         'run in multicore.')
-                
-            if not (hasattr(self, 'cluster') and isinstance(self.cluster, cluster.MultiCore)):                
+            if not (hasattr(self, 'cluster') or not isinstance(self.cluster, cluster.MultiCore)):                
                 self.cluster = cluster.MultiCore(**self.options)
                 self.cluster.nb_core = self.nb_core
+            else:
+                self.cluster.nb_core = self.nb_core
+
 
 
     def clean_previous_results(self,options,p_dirs,folder_name):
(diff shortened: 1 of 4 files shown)


oliviermattelaer:
So I guess that your findings is that when you do 
> self.deamons.clean()
will remove any reference in the threads and that the python garbage collector is then smart enough to tell the system that the thread are effectively dead and should not be counted by the system limit (either put in zombie stated or even removed?). I guess that it might depends of the system obviously. If they are passed in zombie state, you might have a higher limit (but still a limit).

The "issue" with your patch is that you are going to create much more thread than before in some situation. 
The previous code (at least at LO) was trying to reuse previous thread as possible, while you add a lot of .remove() that will force to re-create those at the next submit...

Now if we assume that we do not have limit on zombie/... and that creating thread is cpu free, then we can be even more agressive in the patch applied here.
and apply the 
> self.deamons.clean()
after the end of each wait() 
(and in that case only create one thread per "submit")

The main point of that method will be 
 1. This will likely create even more thread than before but they should be shortlife.
 2. you do not need to call the remove() everywhere in the code (which is designed in case of crash to clean everything properly).
 3. you can add options on the MultiCore class to get back to old behavior (or to force the old behavior in LO case?)

What do you think?

claudioseveri:
Olivier, thank you for looking into this. 
Indeed, my proposal is a bit rough, and was put together in a couple of days to make madgraph run on a shared computer with a tight thread limit. 

I think there are two main lines for the solution to this issue, either:
 1) nb_core threads are created when the main program is opened, and all jobs are submitted to those threads, forever
or:
 2) Threads are created when needed, and closed when their job is done.
 
 I strongly believe 1) is the best option, but I tried to come up with a fix that would not require too drastic changes.
  Right now the code basically does 2), with exceptions, as you point out, but does not close the threads when their job is completed. My patch remains within 2).
 
 
Now to address your points:

> self.deamons.clean() will remove any reference in the threads and the python garbage collector is then smart enough to tell the system that the thread are effectively dead and should not be counted by the system limit
 
This might be the case, but I would definitely **not** count on this, which is why there are calls to .remove() in various places.

 > The "issue" with your patch is that you are going to create much more thread than before in some situation.

I think I only added .remove() where when one can reasonably expect those threads to not be needed anymore (the threads that compile subprocesses are killed after everything is compiled, those that run ajobs when the run is complete, etc)


I can further look into this issue if you want, but I'd be glad to hear your feedback first.

Cheers,
Claudio

oliviermattelaer:
>This might be the case, but I would definitely not count on this, which is why there are calls to .remove() in various places.

actually .remove() does call self.deamons.clean()  (I do not think that the rest of the actions in .remove() has any impact)


>I think I only added .remove() where when one can reasonably expect those threads to not be needed anymore (the threads that compile subprocesses are killed after everything is compiled, those that run ajobs when the run is complete, etc)

I think that within madevent_interface, you are calling .remove() a bit too much.

> I can further look into this issue if you want, but I'd be glad to hear your feedback first.

What I would like to propose to you is that I'm using this branch to implement a solution which does 2 very agressively.
but has also the option to do 1 for the case where we are careful about it. Then I will update this branch such that you can test in your scenario (and I will test other large scan scenario at LO too) to be sure that this is fine before doing the merge

Thanks,

Olivier

oliviermattelaer:
Can you test the latest commit on this branch.
I have tested that for LO scan the number of thread was kept under control (as before actually)
and found that the stoprequest was crucial for doing so.

Thanks in advance,

Olivier

PR #2, opened 2022-06-13
Upc

huashengshao:
The gamma-UPC package is included to enable the UPC simulations at LO (or loop-induced) for exclusive photon-photon processes at pp or p-ion or ion-ion collisions. The usage is just specifying the following parameters in run_card.dat:

2 = lpp1
2 = lpp1
7000.0 = ebeam1
574080.0 = ebeam2
edff or chff = pdlabel
1 = nb_proton1
0 = nb_neutron1
82 = nb_proton2
126 = nb_neutron2

The companion paper is in preparation.


Template/LO/Source/PDF/gammaUPC/ebeam.inc
```diff
@@ -0,0 +1,2 @@
+   ebeam(1)=ebeamMG5(1)/(nb_proton(1)+nb_neutron(1))
+   ebeam(2)=ebeamMG5(2)/(nb_proton(2)+nb_neutron(2))
```
oliviermattelaer, line 2: 1) need to check convention is the same for other ion beam (will do it)
2) why having a .inc here?  This sounds weird to me
3) one need to be carefull with the ebeam/ebeamMG5 name here (I do not like the ebeamMG5  name since both are MG5)

oliviermattelaer (commented):
I think we should 

1) avoid to have to compile the library if not needed by the user.
2) I'm surprised that you do not have any impact parameter added in the run_card.dat
3) Could we add the equation for EDFF and CHFF
4) Note some of the comments are more for me such that I can check them myself

huashengshao:
Hi Olivier, I am done. Please have a look. Cheers, Hua-Sheng

PR #7, opened 2022-06-17
Add vector.inc to many .f files

valassi:
In template/LO, add vector.inc to many fortran files and add -I.. to a few makefiles below Source

See https://github.com/oliviermattelaer/mg5amc_test/issues/16

See https://github.com/madgraph5/madgraph4gpu/issues/455
See https://github.com/madgraph5/madgraph4gpu/issues/458


valassi:
Hi @oliviermattelaer I would like to set your as reviewer but I cannot (ok like this, dont give me too many powers on this repo...).

The rationale for this is essentially to be able to eventually propagate NB_PAGE_MAX and NB_PAGE_LOOP to many fortran files.

A longer explanation is in https://github.com/oliviermattelaer/mg5amc_test/issues/16.

As discussed this morning,  
- hopefully the addition of vector.inc to the files itself should not be controversial (it should not be able to do any harm)
- the only point to discuss is the additions of -I.. to some makefiles: the alternative is that I include '../vector.inc' in the relevant .f files instead, or something similar.
Thanks Andrea

oliviermattelaer:
When testing this, the fortran code does not work anymore (I guess that you have realised that)

```
	    gfortran -w -fPIC -O -fbounds-check -ffixed-line-length-132 -w -c initcluster.f -I../../Source/ -fopenmp
	    vector.inc:1:21:

	    Error: Symbol 'nb_page' at (1) already has basic type of INTEGER
	    vector.inc:3:26:

	    Error: PARAMETER attribute of 'nb_page' conflicts with PARAMETER attribute at (1)
```

Now I would prefer here that we do not use the "-I ../" and indeed include in the "include ../"

valassi:
Hi Olivier, thanks for the comments, I come back to this one.

Points taken, I will update this MR with new code and make sure that:
- I will generate some fortran-only and check that it builds
- I will not add "-I.." to makefiles but rather use "include '../"

Now, the real reason why I need this, as discussed in https://github.com/oliviermattelaer/mg5amc_test/issues/16 and elsewhere, is that I would like to replace all hardcoded numbers for vector sizes by a single parameter. So I will also try to put that in a MR. And eventually also separate two parameters, one for the max size and one for the used size.

Let me clarify this, what I would like to do are two things:

1. Suppose I generate with `output madevent .. --vector_size=16384`. This will hardcode 16384 in two places: in vector.inc as `PARAMETER (NB_PAGE=16384)` and in a variety of other files, for instance as `DOUBLE PRECISION G, ALL_G(16384)`. My first point is that I would like this second type of statement to become `DOUBLE PRECISION G, ALL_G(NB_PAGE)`, but then I also need to include vector.inc in all files. Since Fortran has no include guards (i.e. Fortran INCLUDE is not the same as preprocessor #include), then I need to add the vector.inc each time, and take good care of not breaking other things. This is the point of this MR, and I will try to include all of that here. Is that ok?

2. The second thing that I would like to do is to differentiate between the fact that we statically allocate arrays with 16384 elements, and the fact that we fill in and loop over all 16384 elements. As discussed, the problem is that currently we have no malloc in fortran, and we do static allocations. So the only way out I see in the short term (for an upcoming release) is to have two parameters. This naturally triggers the need to find good names. Currently I call these NB_PAGE_MAX and NB_PAGE_LOOP, but the names are horrible. Maybe we can call the first one VECSIZE (or VECSIZE_MAX?) and the second one VECSIZE_USED (or VECSIZE_LOOP). What do you think about this suggestion? I know we should go for malloc eventually (then we only need one parameter, as in cudacpp), but in the short term I would like to implement this with these two parameters. I can prepare a second MR for this after the first one (or also just have a single one with both).

I think it's becoming urgent that we put these changes upstream. Currently I have these changes as ad-hoc patches in madgraph4gpu, but fo rinstance moving from nuvecMLM to vecMLM was a bit complex because of this... and in any case we cannot give my ptaches to the experiments, it must be upstream.

Let me know... Thanks!
Andrea

valassi:
This is being superseded by https://github.com/mg5amcnlo/mg5amcnlo/pull/23 and can be closed.

I will move my comments there and update them...

PR #17, opened 2022-08-13
Update misc.py to add dummy PACKAGE_INFO in case the VERSION file is not readable

andriish:
Add dummy PACKAGE_INFO in case the VERSION file is not readable etc.

### DIFF
--- a/madgraph/various/misc.py
+++ b/madgraph/various/misc.py
@@ -148,10 +148,14 @@ def get_pkg_info(info_str=None):
                                                   "VERSION"),
                                                   parse_info_str, 
                                                   print_error=False)
-        PACKAGE_INFO = info_dict
-        
-    return info_dict
-
+        if info_dict:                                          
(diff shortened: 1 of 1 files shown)


oliviermattelaer:
Why not, but for which reason that file should not be readable? Is that not a sign of an issue with the installation method?

andriish:
Hi @oliviermattelaer ,

you are partially correct. I've used this patch in a setup with Herwig and the installation tree was moved from the original location and there were problems with the detection of the VERSION file location. That was, however, some versions ago and at this point this patch is quite "pedantic". Exactly as you formulated "Why not".
Maybe one can ad also a warning if the file is not found.
What do you think?

Best regards,

Andrii

oliviermattelaer:
merging done : https://github.com/mg5amcnlo/mg5amcnlo/commit/78908d701ab1142660e309494fc19a040ae035d3

PR #22, opened 2022-12-06
Clear SQUEUE_FORMAT environment variable for job control

APN-Pucky:
I encountered an issue with the slurm cluster mode of MadGraph. 
My jobs got stuck in HOLD and weren't working properly.

A solution to this was to disable the squeue formatting in the environment variable `unset SQUEUE_FORMAT`.
Similarly arguments like `SQUEUE_SORT` exist but should not have an impact.

This PR makes sure that the `SQUEUE_FORMAT` env. var. is unset so that MadGraph can parse the result properly.
Cheers,
Alexander Neuwirth

### DIFF
--- a/madgraph/various/cluster.py
+++ b/madgraph/various/cluster.py
@@ -1729,6 +1729,10 @@ class SLURMCluster(Cluster):
     def control_one_job(self, id):
         """ control the status of a single job with it's cluster id """
         cmd = 'squeue j'+str(id)
+        # Remove incompatible squeue formats 
+        env = os.environ.copy()
+        if "SQUEUE_FORMAT" in env:
+            del env["SQUEUE_FORMAT"]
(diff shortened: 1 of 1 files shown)


oliviermattelaer:
Thanks a lot for this suggestion, this makes a lot of sense.
Will merge this.
The only "issue" is I would like such fix to be implemented first on the long term stable version and not in the main branch.
 looks like the web interface does not allow me to transfer the patch from one branch to the other. So will do it via the command line interface (and therefore I'm not sure who will get credited at the end. Sorry for that).

Thanks,

Olivier

APN-Pucky:
Thanks for the quick reply. I changed the base to `LTS_dev`. 
Sorry for the lazy commit messages, but they should not appear if you squash merge, right?
Otherwise, if this does not work out, I don't care too much about the credits.
Cheers

oliviermattelaer:
That's perfect thanks.

Olivier

PR #5, opened 2022-06-16
Lo vect diagid

oliviermattelaer:
Hi Andrea,

You can merge this branch as soon as you are ready for it.
This branch include change to the amp2 numbering as we discussed before.

For completeness, the main goal was to be able to have the same index for the identification of the amp2 index as for the one related to the definition of the allowed color flow for the event.

I will still improve this branch to be sure/upgrade other mode to have consistent code and/or be sure that everything is fine (like loop-induced/....) but this can be done independently within this branch or to the other ones when this is merged.

PR #23, opened 2022-12-07
First set of vector.inc changes for cudacpp integration

valassi:
Hi @oliviermattelaer,

this is a WIP MR replacing https://github.com/mg5amcnlo/mg5amcnlo/pull/7, which I will close.

I found it cleaner to open a new MR, as I start from a different branch (vecsize) and I target a different branch (vecMLM).

I will add a few comments here - forget the previous MR.

Thanks Andrea

PS (Dec 8) after this morning's discussion I changed the target branch to `gpucpp`. This should become our reference branch for integrating the cudacpp stuff (and eventually portability frameworks), without polluting other 'mainstream' branches of MG5AMC.


valassi:
This MR is meant to ease the integration of the cudacpp plugin. In particular it brings upstream some changes to the Fortran code that until now I had in a custom set of patches (see https://github.com/madgraph5/madgraph4gpu/issues/491).

**NB: this is WIP because it is not yet complete!**

What I plan to do here is the following (suppose I generate with `output madevent .. --vector_size=16384`

1. Rename the parameter NB_PAGE=16384 as parameter VECSIZE_MAX=16384. This is defined in vector.inc. We can discuss the name, but I think we should use something else than NB_PAGE, and I prefer to have a _MAX suffix, see later. This is done.
 
2.  Replace all array definitions with hardcoded sizes `ARRAY(16384)` by definitions in terms of the above, as `ARRAY(VECSIZE_MAX)`. This internally requires that vector.inc is included before all definitions of such arrays, specifically in coupl.inc and/or in various .f files. The largest chunk of changes in this MR come from these modifications to INCLUDE statements. The complication is that Fortran has no include guards (https://github.com/madgraph5/madgraph4gpu/issues/458), so we cannot simply have several inc files inclusing one another. This is also done with the commits so far. Note in particular that I have kept the same Makefiles (so no "-I.." added) and I checked that Fortran seems to build (but please cross check), so this addresses your previous comments to #7.

3. I have noticed that NB_PAGE was used (and so now VECSIZE_MAX is used) in somewhat inconsistent ways in some places. It is a PARAMETER i.e. a global constant, but in some places it is passed as a function argument. This is confusing and a bit error prone. In some places I have replaced VECSIZE_MAX by VECSIZE_MAX_in to indicate that it is an input variable of the function, and not a parameter. This is partly done, but there are many more places. Maybe we should go through them together.

4. The reason I want to call this VECSIZE_MAX is because I need two separate variables, VECSIZE_MAX (the static size of an array at build time) and VECSIZE_USED (the number of elements that are actually filled, and that one should loop over). As discussed, when we move to malloc then we do not need two variables, but for the moment for the upcoming release I would very much prefer to keep this (this is how we derived all the results shown at conferences). Again we can discuss the names, but I would really like to have two variables. This is not yet done, I will probably finish it tomorrow.

Voila, this is all for the moment. Lets discuss tomorrow
Thanks!
Andrea

Template/LO/Source/setrun.f
```diff
@@ -15,6 +15,7 @@ subroutine setrun
       include 'PDF/pdf.inc'
       include 'run.inc'
       include 'alfas.inc'
+      include 'vector.inc'
       include 'MODEL/coupl.inc'
```
oliviermattelaer, line 19: Do no see why this is needed (at least from the diff of the file)
-> ok got it because of coupl.inc 
Can we add a comment about it next to the include?

Template/LO/SubProcesses/genps.f
```diff
@@ -1799,6 +1799,7 @@ double precision function get_channel_cut(p, config)
       include 'nexternal.inc'
       include 'genps.inc'
       include 'maxamps.inc'
+      include 'vector.inc'
       include 'coupl.inc'
```
oliviermattelaer, line 1803: same question here, why is this needed?

Template/LO/SubProcesses/idenparts.f
```diff
@@ -31,6 +31,7 @@ subroutine idenparts(iden_part,itree,sprop,forcebw,prwidth)
 c
 c     Global
 c
+      include 'vector.inc'
       include 'coupl.inc'                     !Mass and width info
```
oliviermattelaer, line 35: same here, why is this needed?

Template/LO/SubProcesses/myamp.f
```diff
@@ -52,6 +52,7 @@ logical function cut_bw(p)
       logical             OnBW(-nexternal:0)     !Set if event is on B.W.
       common/to_BWEvents/ OnBW
       
+      include 'vector.inc'
       include 'coupl.inc'
```
oliviermattelaer, line 56: same

Template/LO/SubProcesses/myamp.f
```diff
@@ -285,6 +286,7 @@ subroutine set_peaks
       double precision stot,m1,m2
       common/to_stot/stot,m1,m2
 
+      include 'vector.inc'
       include 'coupl.inc'
```
oliviermattelaer, line 290: same

Template/LO/SubProcesses/reweight.f
```diff
@@ -1790,6 +1790,7 @@ subroutine update_scale_coupling(p, wgt)
       include 'genps.inc'
       include 'run.inc'
       include 'nexternal.inc'
+      include 'vector.inc'
       include 'coupl.inc'
```
oliviermattelaer, line 1794: same

Template/LO/SubProcesses/setcuts.f
```diff
@@ -8,6 +8,7 @@ SUBROUTINE SETCUTS
 c
       include 'genps.inc'
       include 'nexternal.inc'
+      include 'vector.inc'
       include 'coupl.inc'
```
oliviermattelaer, line 12: same

Template/LO/SubProcesses/setscales.f
```diff
@@ -11,6 +11,7 @@ subroutine set_ren_scale(P,rscale)
 c
       include 'genps.inc'
       include 'nexternal.inc'
+      include 'vector.inc'
       include 'coupl.inc'
```
oliviermattelaer, line 15: same

Template/LO/SubProcesses/setscales.f
```diff
@@ -108,6 +109,7 @@ subroutine set_fac_scale(P,q2factorization)
...
```
oliviermattelaer, line 113: same

Template/LO/SubProcesses/unwgt.f
```diff
@@ -35,6 +35,7 @@ subroutine Zoom_Event(wgt,p)
       double precision xzoomfact
       common/to_zoom/  xzoomfact
       include 'run.inc'
+      include 'vector.inc'
       include 'coupl.inc'
```
oliviermattelaer, line 39: same

madgraph/iolibs/export_v4.py
```diff
@@ -4976,7 +4976,8 @@ def write_vector_size(self, fsock):
         vector_size = banner_mod.ConfigFile.format_variable(vector_size, int, name='vector_size')
         vector_size = max(1, vector_size)
 
-        text = [" integer nb_page\n"," parameter (nb_page=%i)\n" % vector_size]
+        text = [" integer VECSIZE_MEMMAX\n"," parameter (VECSIZE_MEMMAX=%i)\n" % vector_size]
+        text += [" integer VECSIZE_USED\n"," parameter (VECSIZE_USED=VECSIZE_MEMMAX)\n"]
```
oliviermattelaer, line 4981: This sounds ugly... VECSIZE_USED should not be put in constant memory (or I do not see the point of that variable).

madgraph/iolibs/export_v4.py
```diff
@@ -6892,8 +6893,10 @@ def copy_standard_file(self):
         file = open(os.path.join(MG5DIR,\
                               'models/template_files/fortran/rw_para.f')).read()
 
-        includes=["include \'coupl.inc\'","include \'input.inc\'",
-                                              "include \'model_functions.inc\'"]
+        includes=["include \'../vector.inc\'",
+                  "include \'coupl.inc\'",
```
oliviermattelaer, line 6897: This one is less clear, but I guess that it should also be removed correct?

madgraph/iolibs/export_v4.py
```diff
@@ -6984,7 +6992,7 @@ def create_coupl_inc(self):
                 double precision MU_R, all_mu_r%(vec)s
                 common/rscale/ MU_R, all_mu_r
 
-                """   % {'vec': ("(%i)" % max(1,self.vector_size) if self.vector_size else '')}
+                """   % {'vec': ("(VECSIZE_MEMMAX)" if self.vector_size else '')}
```
oliviermattelaer, line 6996: Can we keep the max here ?

madgraph/iolibs/export_v4.py
```diff
@@ -7456,6 +7464,7 @@ def create_couplings_main(self, nb_def_by_file=25):
         fsock.writelines("""logical updateloop
                             common /to_updateloop/updateloop
                             include \'input.inc\'
+                            include \'../vector.inc\'
                             include \'coupl.inc\'
```
oliviermattelaer, line 7468: same (again one which is less clear if we need it or not)

madgraph/iolibs/export_v4.py
```diff
@@ -7521,6 +7530,7 @@ def create_couplings_main(self, nb_def_by_file=25):
                             """)
 
         fsock.writelines("""include \'input.inc\'
+                            include \'../vector.inc\'
                             include \'coupl.inc\'
```
oliviermattelaer, line 7534: same (again one which is less clear if we need it or not)

madgraph/iolibs/export_v4.py
```diff
@@ -7585,6 +7595,7 @@ def create_couplings_main(self, nb_def_by_file=25):
                             'args_dep': ' integer vecid' if self.vector_size else ''
                             })
         fsock.writelines("""include \'input.inc\'
+                            include \'../vector.inc\'
                             include \'coupl.inc\'
```
oliviermattelaer, line 7599: same (again one which is less clear if we need it or not)

madgraph/iolibs/export_v4.py
```diff
@@ -7616,6 +7627,7 @@ def create_couplings_main(self, nb_def_by_file=25):
         #                     double precision mu_r2, as2
         #                     include \'model_functions.inc\'""")
         # fsock.writelines("""include \'input.inc\'
+        #                     include \'../vector.inc\'
         #                     include \'coupl.inc\'
```
oliviermattelaer, line 7631: same (again one which is less clear if we need it or not)

madgraph/iolibs/export_v4.py
```diff
@@ -7640,6 +7652,7 @@ def create_couplings_main(self, nb_def_by_file=25):
                                     include \'mp_coupl.inc\'
                             """%self.mp_real_format)
             fsock.writelines("""include \'input.inc\'
+                                include \'../vector.inc\'
                                 include \'coupl.inc\'
```
oliviermattelaer, line 7656: same (again one which is less clear if we need it or not)

madgraph/iolibs/export_v4.py
```diff
@@ -8000,6 +8013,7 @@ def create_couplings_part(self, nb_file, data, dp=True, mp=False, vec=False):
               parameter  (PI=3.141592653589793d0)
               parameter  (ZERO=0d0)
               include 'input.inc'
+              include '../vector.inc'
               include 'coupl.inc'""")
```
oliviermattelaer, line 8017: same (again one which is less clear if we need it or not)

madgraph/iolibs/template_files/madevent_driver.f
```diff
@@ -71,6 +71,7 @@ Program DRIVER
 c      integer ncols,ncolflow(maxamps),ncolalt(maxamps),ic
 c      common/to_colstats/ncols,ncolflow,ncolalt,ic
 
+      include 'vector.inc'
       include 'coupl.inc'
```
oliviermattelaer, line 75: same

madgraph/iolibs/template_files/madevent_electroweakFlux.inc
```diff
@@ -42,6 +42,7 @@ c     /* ********************************************************* *
 	parameter (eva_pi   = 3.141592653589793d0)
 	parameter (eva_sqr2 = 1.414213562373095d0)
 
+	include '../vector.inc'
 	include '../MODEL/coupl.inc'
```
oliviermattelaer, line 46: same

madgraph/iolibs/template_files/super_auto_dsig_group_v4.inc
```diff
@@ -114,23 +114,23 @@ C     Cannot make a selection with all PDFs to zero, so we return now
...
 c     INPUT (VIA COMMAND BLOCK)
 c       SELPROC 
 c       SUMPROB
 c     INPUT
-c       nb_page (number of weight to update)
+c       VECSIZE_USED (number of weight to update)
 c     INPUT/OUTPUT
-c       WGTS(nb_page) #multiplied by the associated jacobian      
+c       WGTS(VECSIZE_USED) #multiplied by the associated jacobian      
 c
```
oliviermattelaer, line 128: typo here (not introduced by you)  it should be wgt and not wgts

models/template_files/fortran/printout.f
```diff
@@ -9,6 +9,7 @@
       subroutine printout
       implicit none
 
+      include '../vector.inc'
       include 'coupl.inc'
```
oliviermattelaer, line 13: same

oliviermattelaer (commented):
So this sounds good, mainly cosmetic change (adding comment).
The only part that is puzzling me is that they are no additional feature allowing to set vecsize_used by the user. It is currently hardcoded to the maximal value  in constant memory. Do you want me to add such feature via the run_card? (note that this will still force a recompilation of the code since any change to the run_card requires a (very small) compilation of the code.

valassi:
H

> So this sounds good, mainly cosmetic change (adding comment). The only part that is puzzling me is that they are no additional feature allowing to set vecsize_used by the user. It is currently hardcoded to the maximal value in constant memory. Do you want me to add such feature via the run_card? (note that this will still force a recompilation of the code since any change to the run_card requires a (very small) compilation of the code.

Hi Olivier, thanks for the feedback!

No about the hardcoding of vesize_used I am doing this in later patches
- not even done in MR #24 (there I think it just becomes a DATA instead of a PARAMETER, but no one changes)
- the runtime choice is done later at runtime in a patch that is still in my madgraph4gpu...

I am trying to go step by step... Here for MR #23 the main point was fixing the includes and separating the two varaibles

I will look at your cosmetic changes in detail

Template/LO/Source/setrun.f
valassi: Yes here vector.inc was included because of coupl.inc. I have added the comment inside coupl.inc actually. In generated code it looks like this
https://github.com/madgraph5/madgraph4gpu/blob/79a0b8f71f3a9e2922494493250c5fd8985706c0/epochX/cudacpp/gg_tt.mad/Source/MODEL/coupl.inc#L7
```
C     NB: VECSIZE_MEMMAX is defined in vector.inc
C     NB: vector.inc must be included before coupl.inc
C     

      DOUBLE PRECISION G, ALL_G(VECSIZE_MEMMAX)
      COMMON/STRONG/ G, ALL_G
```

But I agree that I can add extra comments. How about changing the various
```
      include 'vector.inc'
```
into
```
      include 'vector.inc' ! needed by coupl.inc (defines VECSIZE_MEMMAX)
```
Looks better?

Template/LO/SubProcesses/idenparts.f
valassi: same as above, vector.inc is needed by coupl.inc

Template/LO/SubProcesses/myamp.f
valassi: same as above, vector.inc is needed by coupl.inc

Template/LO/SubProcesses/myamp.f
valassi: same as above, vector.inc is needed by coupl.inc

Template/LO/Source/setrun.f
oliviermattelaer: yes wonderfull. Thanks a lot

madgraph/iolibs/export_v4.py
valassi: same as above, vector.inc is needed by coupl.inc

madgraph/iolibs/template_files/super_auto_dsig_group_v4.inc
valassi: Ok I will change that

models/template_files/fortran/printout.f
valassi: same as above, vector.inc is needed by coupl.inc

Template/LO/SubProcesses/reweight.f
valassi: same as above, vector.inc is needed by coupl.inc

Template/LO/SubProcesses/setcuts.f
valassi: same as above, vector.inc is needed by coupl.inc

Template/LO/SubProcesses/setscales.f
valassi: same as above, vector.inc is needed by coupl.inc

Template/LO/SubProcesses/setscales.f
valassi: same as above, vector.inc is needed by coupl.inc

Template/LO/SubProcesses/unwgt.f
valassi: same as above, vector.inc is needed by coupl.inc

madgraph/iolibs/export_v4.py
valassi: same as above, vector.inc is needed by coupl.inc

madgraph/iolibs/export_v4.py
valassi: hm I need to think about that... what is the max used for?

madgraph/iolibs/export_v4.py
valassi: same as above, vector.inc is needed by coupl.inc

madgraph/iolibs/export_v4.py
valassi: same as above, vector.inc is needed by coupl.inc

madgraph/iolibs/export_v4.py
valassi: same as above, vector.inc is needed by coupl.inc

madgraph/iolibs/export_v4.py
valassi: same as above, vector.inc is needed by coupl.inc

madgraph/iolibs/template_files/madevent_driver.f
valassi: same as above, vector.inc is needed by coupl.inc

madgraph/iolibs/export_v4.py
valassi: yes it is ugly for the moment, but this is an interim patch, just the first step to fix the includes...

in the next MR #24 the parameter becomes a DATA and is moved around nicely

and in my madgraph4gpu patches (since a long time...) the DATA is read from file when you do "madevent < input_file", see https://github.com/madgraph5/madgraph4gpu/blob/79a0b8f71f3a9e2922494493250c5fd8985706c0/epochX/cudacpp/gg_tt.mad/SubProcesses/P1_gg_ttx/driver.f#L105

```
      write(*,*) 'Enter #events in a vector loop (max=',VECSIZE_MEMMAX,',)'
      read(*,*) VECSIZE_USED
...
      write(*,'(a16,i6)') ' VECSIZE_USED = ', VECSIZE_USED
      if( VECSIZE_USED.gt.VECSIZE_MEMMAX .or. VECSIZE_USED.le.0 ) then
        write(*,*) 'ERROR! Invalid VECSIZE_USED = ', VECSIZE_USED
        STOP
      endif
```

This is a typical input file for madevent in my tests
https://github.com/madgraph5/madgraph4gpu/blob/79a0b8f71f3a9e2922494493250c5fd8985706c0/epochX/cudacpp/tmad/logs_ggtt_mad/log_ggtt_mad_d_inl0_hrd0.txt#L110
```
*** (2-none) EXECUTE CMADEVENT_CUDACPP x1 (create events.lhe) ***
--------------------
+1 ! Fortran bridge mode (CppOnly=1, FortranOnly=0, BothQuiet=-1, BothDebug=-2)
8192 ! Number of events in a single C++ or CUDA iteration (VECSIZE_USED)
8192 1 1 ! Number of events and max and min iterations
0.000001 ! Accuracy (ignored because max iterations = min iterations)
0 ! Grid Adjustment 0=none, 2=adjust (NB if = 0, ftn26 will still be used if present)
1 ! Suppress Amplitude 1=yes (i.e. use MadEvent single-diagram enhancement)
0 ! Helicity Sum/event 0=exact
1 ! Channel number (1-N) for single-diagram enhancement multi-channel (NB used even if suppress amplitude is 0!)
--------------------
```

madgraph/iolibs/export_v4.py
oliviermattelaer: I do not fully remember to be honest it was introduce to be able to catch specific case where vectorization is not used.

madgraph/iolibs/export_v4.py
oliviermattelaer: So in production, I would stay that we need to go back to the normal parameter setting for input_app.txt
since otherwise we will need to change a lot of python file that does handle the creation of those many files.
(in various ways).

This especially makes sense, since you want that parameter to be identical for all channel of integration.
The point of that file is to handle parameter which are channel dependent.

The parameter "Fortran bridge mode" should also be moved to the run_card.dat.

And for the C++ we should (really) test the following setup for the input_app.txt

*** (2-none) EXECUTE CMADEVENT_CUDACPP x1 (create events.lhe) ***
--------------------
+1 ! Fortran bridge mode (CppOnly=1, FortranOnly=0, BothQuiet=-1, BothDebug=-2)
8 ! Number of events in a single C++ or CUDA iteration (VECSIZE_USED)
1000 7 3 ! Number of events and max and min iterations
0.1 ! Accuracy (ignored because max iterations = min iterations)
2 ! Grid Adjustment 0=none, 2=adjust (NB if = 0, ftn26 will still be used if present)
1 ! Suppress Amplitude 1=yes (i.e. use MadEvent single-diagram enhancement)
0 ! Helicity Sum/event 0=exact
1 ! Channel number (1-N) for single-diagram enhancement multi-channel (NB used even if suppress amplitude is 0!)
--------------------

Since this is what is going to be used in practise. (we can change the "1000" to be a multiple of 8 obviously)

madgraph/iolibs/export_v4.py
valassi: Ok I wil try to have a look

madgraph/iolibs/export_v4.py
valassi: Hi Olivier, Ok I see your points. I agree, in production we will need to put those much more hardcoded somewhere, eg in the run card (but then also VECSIZE_MEMMAX can be hardcoded back to lower numbers)... I will think about it.

One option otherwise for a runtime parameter is the parameter card?

Yes the fortran bridge should go in a runcard or a parameter card. It depends if you want to rebuild or not.

As for the value=8, I do routinely test 32. I think this is a safer bet than 8, with little problems. Or at least 16 to allow also float AVX512.

madgraph/iolibs/export_v4.py
valassi: Hi Olivier, I am not really sure what the logic was, but I changed it so that it is "1" hardcoded if vector_size <=1,
https://github.com/mg5amcnlo/mg5amcnlo/pull/23/commits/066fb101243a81179b29d6144306ae94de971cdf
Hope this can be resolved?

madgraph/iolibs/template_files/super_auto_dsig_group_v4.inc
valassi: Changed here https://github.com/mg5amcnlo/mg5amcnlo/pull/23/commits/dd98f006ce759c1d4d93aaeeaa20ecf080f6f9e0
Can be resolved

Template/LO/Source/setrun.f
valassi: Done here https://github.com/mg5amcnlo/mg5amcnlo/pull/23/commits/49100837bd872e4f4abe104ddb0c1a974939741f

Can be resolved

valassi:
Hi @oliviermattelaer I have made the few changes you suggested.

If you are ok with this now, please merge it. Then we can focus on the second one vecsize2 (in parallel to the other lo_vect_diag, lets do thes etwo independently)

valassi:
@oliviermattelaer I have now cleaned up this first vecsize #23 and the second vecsize2 #24. They are essentially what they were before I attempted any changes due to lo_vect_diag #5, which we agreed to merge.

As you have already made comments on this first patch #23, and I have implemented them, I will just go ahead and merge this (into a dedicated "gpucpp" branch anyway...). This will make it easier for you to review the second set pf patches in #24.

Hope this is ok! Thanks Andrea

valassi:
Self merging.

PR #24, opened 2022-12-10
Second set of vector.inc changes for cudacpp integration (make VECSIZE_USED a function argument)

valassi:
Hi @oliviermattelaer @roiser this is a (WIP, for now) second set of vector.inc changes for cudacpp integration.

It is a followup to #23 and it includes it internally (as that has not been merged yet).

It is addressing the part that was still missing in #23, namely it transforms VECSIZE_USED from a global constant parameter into a local variable that is passed through the code as a function argument. This is what OLivier had suggested during our chat on Thursday morning.

More details from the log of https://github.com/mg5amcnlo/mg5amcnlo/commit/fc4dfd92529928afc4d92c21eaf589aa88cd7f23
```
This is how VECSIZE_USED is currently passed throughout a Fortran-only application:
- in driver.f, program DRIVER reads VECSIZE_USED from user inputs
- in driver.f, program DRIVER calls SAMPLE_FULL with argument VECSIZE_USED
 > in dsample.f, subroutine SAMPLE_FULL calls SAMPLE_INIT (in dsample.f) with argument VECSIZE_USED
  == in dsample.f, subroutine SAMPLE_INIT uses argument VECSIZE_USED
 > in dsample.f, subroutine SAMPLE_FULL calls SELECT_GROUPING with argument VECSIZE_USED
  == in auto_dsig.f, subroutine SELECT_GROUPING uses argument VECSIZE_USED
 > in dsample.f, subroutine SAMPLE_FULL calls DSIG_VEC with argument VECSIZE_USED
  > in auto_dsig.f, subroutine DSIG_VEC calls UPDATE_SCALE_COUPLING_VEC with argument VECSIZE_USED
   == in reweight.f, subroutine UPDATE_SCALE_COUPLING_VEC uses argument VECSIZE_USED
  > in auto_dsig.f, subroutine DSIG_VEC calls DSIG_PROC_VEC with argument VECSIZE_USED
   > in auto_dsig.f, subroutine DSIG_PROC_VEC calls DSIG1_VEC with argument VECSIZE_USED
    > in auto_dsig1.f, subroutine DSIG1_VEC calls SMATRIX1_MULTI with argument VECSIZE_USED
     == in auto_dsig1.f, subroutine SMATRIX1_MULTI uses argument VECSIZE_USED
```

This is almost complete, actually the last log says it is complete, but I will change it and amend the log with a force push, so I leave it in WIP for now. The only things that I would like to change are the following
- The comments in vector.inc refer to the case when the code is generated in vector mode. When vector mode is not used (i.e. when VECSIZE_MEMMAX is 1), I think that it would be better to use different comments. Or at least, the case VECSIZE_MEMMAX=1 should be mentioned in the comments.
- The initialization of VECSIZE_USED is now a DATA block in driver.f. This is how I plan to use it when the cudacpp integration is complete (i.e. when my next set of patches is ported upstream). But for the moment, this code without cudacpp would make more sense with a PARAMETER definition in driver.f. This is esepcially the case for VECSIZE_MEMMAX=1 above.
- I will also leave this in WIP until the previous #23 is merged, so it becomes clearer what this one adds.

Anyway, these are almost cosmetic changes. The code now builds (and hopefully runs) also in fortran only mode, both in vector and no vector mode.

I have an upcoming MR of madgraph4gpu that is based on this one, I will add the reference when I create it.

### DIFF
--- a/Template/LO/Source/dsample.f
+++ b/Template/LO/Source/dsample.f
@@ -1,16 +1,17 @@
-      subroutine sample_full(ndim,ncall,itmax,itmin,dsig,ninvar,nconfigs)
+      subroutine sample_full(ndim,ncall,itmax,itmin,dsig,ninvar,nconfigs,VECSIZE_USED)
 c**************************************************************************
 c     Driver for sample which does complete integration
 c     This is done in double precision, and should be told the
 c     number of possible phasespace choices.
 c     Arguments:
-c     ndim       Number of dimensions for integral(number or random #'s/point)
-c     ncall      Number of times to evaluate the function/iteration
-c     itmax      Max number of iterations
-c     itmin      Min number of iterations
-c     ninvar     Number of invarients to keep grids on (s,t,u, s',t' etc)
-c     nconfigs   Number of different pole configurations 
-c     dsig       Function to be integrated
+c     ndim           Number of dimensions for integral(number or random #'s/point)
+c     ncall          Number of times to evaluate the function/iteration
+c     itmax          Max number of iterations
+c     itmin          Min number of iterations
+c     dsig           Function to be integrated
+c     ninvar         Number of invarients to keep grids on (s,t,u, s',t' etc)
+c     nconfigs       Number of different pole configurations 
+c     VECSIZE_USED   Number of events in parallel out of VECSIZE_MEMMAX
 c**************************************************************************
       implicit none
       include 'genps.inc'
@@ -19,6 +20,7 @@ c
 c Arguments
 c
       integer ndim,ncall,itmax,itmin,ninvar,nconfigs
+      integer VECSIZE_USED
       external         dsig
       double precision dsig
 c
@@ -141,7 +143,7 @@ c-----
 C     Fix for 2>1 process where ndim is 2 and not 1
       ninvar = max(2,ninvar)
 
-      call sample_init(ndim,ncall,itmax,ninvar,nconfigs)
+      call sample_init(ndim,ncall,itmax,ninvar,nconfigs,VECSIZE_USED)
       call graph_init
       do i=1,itmax
          xmean(i)=0d0
@@ -374,7 +376,7 @@ c
       ncall = ncall*4 ! / 2**(itmax-2)
       write(*,*) "Starting w/ ncall = ", ncall
       itmax = 8
-      call sample_init(ndim,ncall,itmax,ninvar,nconfigs)
+      call sample_init(ndim,ncall,itmax,ninvar,nconfigs,VECSIZE_USED)
       do i=1,itmax
          xmean(i)=0d0
          xsigma(i)=0d0
@@ -652,7 +654,7 @@ c     $     ntot/1000,'</th><th align=right>',teff,'</th></tr>'
 
 
 
-      subroutine sample_init(p1, p2, p3, p4, p5)
+      subroutine sample_init(p1, p2, p3, p4, p5, VECSIZE_USED)
 c************************************************************************
 c     Initialize grid and random number generators
 c************************************************************************
@@ -668,7 +670,7 @@ c
 c     Arguments
 c
       integer p1, p2, p3, p4, p5
-
+      integer VECSIZE_USED
 c
 c     Local
 c
--- a/Template/LO/SubProcesses/cuts.f
+++ b/Template/LO/SubProcesses/cuts.f
@@ -256,11 +256,15 @@ c               fixed_ren_scale=.true.
 c               call set_ren_scale(P,scale)
 c            endif
 c         endif
-         
+
+c     If scale is fixed, update G-dependent couplings for VECSIZE_MEMMAX events
+c     This is called only once in the application (FIRSTTIME=.true.)
+c     Only VECSIZE_USED events are needed, but this variable is unknown here
+c     Using VECSIZE_MEMMAX is correct and has a negligible performance overhead
 
          if(fixed_ren_scale) then
             G = SQRT(4d0*PI*ALPHAS(scale))
-            do i =1, VECSIZE_USED
+            do i =1, VECSIZE_MEMMAX ! no need to use VECSIZE_USED here
                call update_as_param(i)
             enddo
          endif
--- a/madgraph/iolibs/export_v4.py
+++ b/madgraph/iolibs/export_v4.py
@@ -6334,7 +6334,7 @@ class ProcessExporterFortranMEGroup(ProcessExporterFortranME):
                 "IF(IPROC.EQ.%(num)d) DSIGPROC=DSIG%(num)d(P1,WGT,IMODE) ! %(proc)s" % data
                 )
             call_dsig_proc_lines_vec.append(\
-                "IF(IPROC.EQ.%(num)d) CALL DSIG%(num)d_VEC(ALL_P1,ALL_XBK, ALL_Q2FACT,ALL_CM_RAP,ALL_WGT,IMODE,ALL_OUT) ! %(proc)s" % data
+                "IF(IPROC.EQ.%(num)d) CALL DSIG%(num)d_VEC(ALL_P1,ALL_XBK,ALL_Q2FACT,ALL_CM_RAP,ALL_WGT,IMODE,ALL_OUT,VECSIZE_USED) ! %(proc)s" % data
                 )
 
         replace_dict['call_dsig_proc_lines'] = "\n".join(call_dsig_proc_lines)
--- a/madgraph/iolibs/export_v4.py
+++ b/madgraph/iolibs/export_v4.py
@@ -4975,9 +4975,24 @@ class ProcessExporterFortranME(ProcessExporterFortran):
             vector_size = 1
         vector_size = banner_mod.ConfigFile.format_variable(vector_size, int, name='vector_size')
         vector_size = max(1, vector_size)
-
-        text = [" integer VECSIZE_MEMMAX\n"," parameter (VECSIZE_MEMMAX=%i)\n" % vector_size]
-
+        text=["""C
+C Offloading ME to GPUs or vectorized C++ needs a vector API.
+C Fortran arrays in this API can hold up to VECSIZE_MEMMAX events.
+C These arrays are statically allocated at compile time.
+C The constant value of VECSIZE_MEMMAX is fixed at codegen time
+C (output madevent ... --vector_size=<VECSIZE_MEMMAX>).
+C
+C While the arrays can hold up to VECSIZE_MEMMAX events,
+C only VECSIZE_USED (<= VECSIZE_MEMAMX) are used in Fortran loops.
+C The value of VECSIZE_USED can be chosen at runtime
+C (typically 8k-16k for GPUs, 16-32 for vectorized C++).
+C
+C NB: THIS FILE CANNOT CONTAIN #ifdef DIRECTIVES
+C BECAUSE IT DOES NOT GO THROUGH THE CPP PREPROCESSOR
+C (see https://github.com/madgraph5/madgraph4gpu/issues/458).
+C
+      INTEGER VECSIZE_MEMMAX
+      PARAMETER (VECSIZE_MEMMAX=%i)""" % vector_size]
         fsock.writelines(text)
         return vector_size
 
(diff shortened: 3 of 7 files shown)


valassi:
Hi @oliviermattelaer @roiser this second set of upstream patches in mg5amcnlo is now also complete.

It includes MR #23, which should be merged first. With respect to that one, essentially it cleans up VECSIZE_USED and makes it a function argument, rather than a global parameter or common.

My second madgraph4gpu "lhe" patch is based on it, https://github.com/madgraph5/madgraph4gpu/pull/562

En passant, note that the large speedup I observed in fortran between nuvecMLM and vecMLM, which is the basis for #23 and this #24, is now understood. Somehow, Olivier's latest changes improved OMP multithreading, so the factor x4 was coming from my 4-core node. I have now disabled OMP multithreading by default in madgraph4gpu, https://github.com/madgraph5/madgraph4gpu/issues/561

Let me know if you have comments, otherwise I would self merge both #23 and #24. They will go into a specific gpucpp branch anyway, not into vecMLM.

Thanks Andrea

oliviermattelaer:
Ok this is problematic for me to do a review since the diff is based on the wrong branch (so I need to recheck all change propose in the other merge first). 

IN launchpad we can handle that by requiring that the other MR to be acccepted first and then see only the additional diff.
I do not find such option here, and looks like I can not change the branch target to the other one since you did not push it within mg5amcnlo ... Do you have a workaround ?

Olivier

valassi:
Hi Olivier, yes you are right it is easier if we accept #23 first. Otherwise I should have created separate target branches... maybe easier to just get #23 out of the way first?

valassi:
I also added here the few changes you suggested in #23. Once #23 is merged, it will become clearer what are the changes in this #24

valassi:
Hi Olivier, as discussed I have cleaned up my two patches vecsize #23 and this vecsize2 #24 taking into account that we agreed to drop lo_vect_diag. Since you had already given comments on the first patch, I just went ahead and merged it (into "gpucpp"). This is the second set of patches, which now should be much more readable. 

Indeed it is very simple. The point was just to clean up the code to make VECSIZE_USED a function argument, as at some point you had suggested amongst various options I proposed.

Could you give your feedback on this one if you have any? Then I can merge it and we can move on from there.

Thanks!
Andrea

Template/LO/Source/dsample.f
```diff
@@ -1,16 +1,17 @@
...
-c     nconfigs   Number of different pole configurations 
-c     dsig       Function to be integrated
+c     ndim           Number of dimensions for integral(number or random #'s/point)
+c     ncall          Number of times to evaluate the function/iteration
+c     itmax          Max number of iterations
+c     itmin          Min number of iterations
+c     dsig           Function to be integrated
+c     ninvar         Number of invarients to keep grids on (s,t,u, s',t' etc)
+c     nconfigs       Number of different pole configurations 
+c     VECSIZE_USED   Number of events in parallel out of VECSIZE_MEMMAX
```
oliviermattelaer, line 14: Can you specified what parallel means here
Specifically, it should be specified if this is the lock-step size or just a packet size which should not be in lock-step.

madgraph/iolibs/export_v4.py
```diff
@@ -4975,10 +4975,29 @@ def write_vector_size(self, fsock):
...
+C
+C While the arrays can hold up to VECSIZE_MEMMAX events,
+C only VECSIZE_USED (<= VECSIZE_MEMAMX) are used in Fortran loops.
+C The value of VECSIZE_USED can be chosen at runtime
+C (typically 8k-16k for GPUs, 16-32 for vectorized C++).
+C
+C NB: THIS FILE CANNOT CONTAIN #ifdef DIRECTIVES
+C BECAUSE IT DOES NOT GO THROUGH THE CPP PREPROCESSOR
+C (see https://github.com/madgraph5/madgraph4gpu/issues/458).
+C
```
oliviermattelaer, line 4998: Here as well we need to add a discussion on the lock-size size.

oliviermattelaer (approved):
Hi Andrea,

I still think that this is not enough. We should have 
third parameter
- we do have VECSIZE_MEMMAX for memory handling
- we do have VECSIZE_USED for memory handling

We need VECSIZE_LOCK to say how many of those VECSIZE_USED need to be generated such that lock-step is possible (that number can not be large, so 32 for GPU if it is much larger, I need to slow down the first iteration of madevent to ensure that the phase-space integration do cover the full phase-space.

I guess this is ok to merge this and then work on the merge branch to change some vecsize_used to 


do i1=0, vecsize_used/vecsize_lock
do i2 = 0, vecsize_lock
   i = i1 * vecsize_lock + i2
  # do here all selection that depends on the lock size
enddo
enddo

This will then also require that some metadata to take a size since they can be different for each "i2" value

Template/LO/SubProcesses/cuts.f
```diff
@@ -256,11 +256,15 @@ LOGICAL FUNCTION PASSCUTS(P)
...
+c     If scale is fixed, update G-dependent couplings for VECSIZE_MEMMAX events
+c     This is called only once in the application (FIRSTTIME=.true.)
+c     Only VECSIZE_USED events are needed, but this variable is unknown here
+c     Using VECSIZE_MEMMAX is correct and has a negligible performance overhead
 
          if(fixed_ren_scale) then
             G = SQRT(4d0*PI*ALPHAS(scale))
-            do i =1, VECSIZE_USED
+            do i =1, VECSIZE_MEMMAX ! no need to use VECSIZE_USED here
                call update_as_param(i)
```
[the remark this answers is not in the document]
valassi: Hi, if I remember correctly... you are completely correct that VECSIZE_USED would be more appropriate here. The problem is that, if we have this as an argument to pass from function to function and not as a common, then I had found it difficult to pass it here. But I can try again if you want.

This is in the passcut function, which is called in two places only IIUC

1. In dsigproc (auto_dsig.f), where however I see
```
      DSIGPROC=0D0
C     not needed anymore ... can be removed ... set for debugging only
C        
      IF (.NOT.PASSCUTS(P1)) THEN
        STOP 1
      ENDIF
``` 
Maybe this can be removed?...

2. In sample_full from dsample.f, where indeed vecsize_used is available, so I can add that trivially

3.  In addition there are two more calls to passcuts, both commented out
- one in dsigproc_vec from auto_dsig.f
- one in gen_mom from genps.f

Let's say: if you confirm that the two commented out lines can be removed (or kepyt commented out), and that also the dsig call can be commented out, then I can add vecsize_used to the passcuts function signature as an argument, and use it instead of vecsize_memmax. Should I do that?

Otherwise, I do not think that you get a segfault, just because in practice update_as_param(i) and the coup2(i) that it calls internally are both setting a variable(i), with no input variable at all. The memory is allocated up to vecsize_memmax, so you just end up creating a vector that is too long, longer than is needed, but the performance penalty is really negligible as I think this is done once, not for each event. Ugly, but it would work.

What do you prefer, should I change the passcuts signature adding vecsize_used and comment out passcuts in sigproc?

valassi:
> Hi Andrea,
> 
> I still think that this is not enough. We should have third parameter
> 
>     * we do have VECSIZE_MEMMAX for memory handling
> 
>     * we do have VECSIZE_USED for memory handling
> 
> 
> We need VECSIZE_LOCK to say how many of those VECSIZE_USED need to be generated such that lock-step is possible (that number can not be large, so 32 for GPU if it is much larger, I need to slow down the first iteration of madevent to ensure that the phase-space integration do cover the full phase-space.
> 
> I guess this is ok to merge this and then work on the merge branch to change some vecsize_used to
> 
> ```
> do i1=0, vecsize_used/vecsize_lock
> do i2 = 0, vecsize_lock
>    i = i1 * vecsize_lock + i2
>   # do here all selection that depends on the lock size
> enddo
> enddo
> ```
> 
> This will then also require that some metadata to take a size since they can be different for each "i2" value

Hi Olivier, thanks!

Maybe better discuss this in person? I think we briefly discussed this in the past, I am not sure we disagree, maybe we do not understand each other.

What I have in mind is that the Fortran does not need to know the locksize step that is actually used inside cudacpp. The Fortran should assume that up to VECSIZE_USED can be needed for lockstep, and therefore all of those should be provided. 

To be clear, in the example of vectorization: whether we use SSE4 or AVX2 or AVX512 is something that currently is only visible inside cudacpp, and I would keep it this way. There is no need to expose this to the Fortran (and it would even be relatively complex/cumbersome to do so).

In practice, what I suggest is something like the following, in production:
- For CPUs and CPP, we have VECSIZE_MEMMAX=32 and VECSIZE_USED=32. The Fortran should always assume that 32 events at the same time go in parallel. Then whether cudacpp actually does it or not, it's up to cudacpp to decide.
- For GPUs, *for the moment* we always have something like VECSIZE_MEMMAX=32768 and VECSIZE_USED=32768 (alternative: both 16384). Fortran should assume that these are the numbers of events to process in parallel on the GPU kernels.  

Note that **in production** above I suggest that VECSIZE_MEMMAX and VECSIZE_USED are always the same! So you might wonder, why do I make all this fuss about having two spearate ones? The reason is only that, **for development**, now I am building only one Fortran that I use both with cuda and cpp (this may change if we use different directories: https://github.com/madgraph5/madgraph4gpu/issues/502). So **for development** I am always setting VECSIZE_MEMMAX=16384, whether for GPU or CPU, and then I use VECSIZE_USED to choose at runtime between the two.

Maybe eventually we can even go back to a single parameter, but currently there are just too many "chantiers ouverts" and using this trick is easy and effective.

In all this, again I do not see the need for the lock size to appear in the Fortran. But yes I can add that as a comment!

Final point: *eventually*, when we go to smaller kernels, one GPU thread per helicity etc, we can probably reduce the VECSIZE_USED on GPU from 16384 to much fewer. But *now* this is unavoidable...

Does this make sense? Let's have a chat on the phone otherwise
Thanks!
Andrea

madgraph/iolibs/export_v4.py
valassi: Sure I can add a comment that the actual SIMD lockstep size can be lower than VECSIZE_USED

Template/LO/Source/dsample.f
valassi: Sure I can add a comment that the actual SIMD lockstep size can be lower than VECSIZE_USED. 

The VECSIZE_USED is just the parallelism size that the bridge understands, it does the same on all those events. Then internally it may take several CPU loops, or several SIMD vectors, or also several GPU warps, to achieve that.

Template/LO/SubProcesses/cuts.f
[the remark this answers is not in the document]
valassi: Urgh now I see another thing that I did not like. Yes I can pass VECSIZE_USED to passcuts, but this is really ugly because actually it is only used once. 

The problem is that the first call to passcuts is where the couplings are set (depending on scale)! This is not particularly nice. It would probably be better to do it outside passcuts?

valassi:
Hi @oliviermattelaer I created a small patch to address your comments 
https://github.com/mg5amcnlo/mg5amcnlo/pull/31/files
Is it better? Thanks Andrea

Template/LO/SubProcesses/cuts.f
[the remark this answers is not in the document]
oliviermattelaer: One think that is also ugly is that I would anyway need to set one variable as a common block since 
VECSIZE_USED  needs to be part of the run_card.dat and that those parameter are for most of them --if not all--
have their definition in run.inc and their value in run_card.inc  which is include/evaluated during the initialization.

So for the moment, I was thinking to set for the production code
VECSIZE_USED_USER part of the common block
and ensure that in production
VECSIZE_USED_USER = VECSIZE_USED 
and ensure that
VECSIZE_MEMMAX>= VECSIZE_USED_USER 
(so force a recompilation if that one is increase but do not recompile if it is decrease)

The advantages of having VECSIZE_USED_USER different of  VECSIZE_USED is that you can have as much hack as you need within the code and have that value changed at anytime, which I would not like for VECSIZE_USED_USER.
But maybe you do not need that.

For the passcuts, I agree that that one is quite ugly to have. If this is for initialization maybe using VECSIZE_USED_USER is ok. For the pascuts, I would say that this is important to be able to call it from anywhere in the code.

But indeed, it might be a good idea to move that part of the code outside of the cuts function where it does not make that much sense. ( will look for that in "my" other work in progress branch)

PR #27, opened 2023-01-04
Customization frameworks: custom cuts/scale/bias, custom entry in the run_card, default way to set default run_card entry

oliviermattelaer:
This thread started due to this request:
https://bugs.launchpad.net/mg5amcnlo/+bug/2001583

The idea is to auto-detect the type of missing entry in the run_card (but to check that the user does not use a LO card for a NLO one, or the opposite).

To make this more useful: 
 - the new entry are automatically defined in run.inc and added to a common block
 - a new entry in the run_card (custom_fct) allows to overwrite some fortran function, such that such parameters can be used within custom functions of cuts/scale/bias 
 - add file to setup default value for the run_card (both lo and nlo) allowing a user to set such entry by default for various runs

status tracking:
 - [x] able to detect the type and to force it via prefix like list_str_, str_, int_, ...
 - [x] able to add metadata with name suffix like <include=False><fortran_name=XXX>  
 - [x] support list  (at least partly dictionary too)
 - [x] security to avoid LO/NLO mismatch
 - [x] automatic adding the missing variable to run.inc (default file, possible to change it to another file)
 - [x] previous definition are automatically cleaned in run.inc (other files might not be cleaned automatically)
 - [x] unittest (5: test_guess_type test_guess_entry_fromname test_add_unknown_entry test_add_definition test_autodef_nomissmatch)
 - [x] basic check that it work in a real run (note no acceptance test associated to this feature. test done manually)
 - [x] set new variable in common block
 - [x] allow to customize a predefined set of user-hook via the run_card (+ external files) at LO (including cuts and scale)
 - [x] allow to customize a predefined set of user-hook via the run_card (+ external files) at NLO for cuts
 - [x] allow to customize a predefined set of user-hook via the run_card (+ external files) at NLO for scale
 - [x] allow to customize a predefined set of user-hook via the run_card (+ external files) at NLO for bias
 - [x] allow to have a file input/run_card_lo_default.dat that overwrite default value when the card is generated
 - [x] allow to have a file input/run_card_lo_default.dat that overwrite default value when the card is generated


oliviermattelaer:
added a new FAQ related to implementing cuts, since this simplifies the procedure:
https://answers.launchpad.net/mg5amcnlo/+faq/3323

oliviermattelaer:
Based on the above work, I thought it will be nice to extend the allowed customization even further to actually add something really interesting for MG5aMC (and not only for hack/plugin)

I have added one more functionality (only at Lo but conceptually compatible with NLO).
The idea is to have a new entry custom_fcts in the run_card.dat which accepts a list of path to file.
Those files can contains one or more functions (only a specific list of function are supported) that would then over-write the existing function within the code.
Two cases are useful for LO:
 - adding new cuts to the code (via the function: dummy_cuts)
 - defining a new dynamical scale (via the function: user_dynamical_scale)
even the list of supported function is longer (all the functions of dummy_fct.f) which was designed for PLUGIN

(Note a dedicated unittest testing the functionality has been added)

So I plan to add two additional things:
 - [ ] put new variable within a common_block (default would be user_customization) (This should allow very simple implementation of new cuts with those two new features) 
 - [ ] Some restructuring of some NLO file to have the LO equivalent of dummy_fct.f (i.e. a file designed for customization) and allow to
       -  define new cuts
       -  define dynamical scale formula
       -  define bias function

Template/LO/Cards/run_card.dat
```diff
@@ -79,11 +79,12 @@ $ckkw
 			     ! 2 is new strategy (using only the denominator)
 $psoptim			     
 #*********************************************************************
-# Generation bias, check the wiki page below for more information:   *
+# Generation bias and customization (custom cuts/scale/...)          *
 #  'cp3.irmp.ucl.ac.be/projects/madgraph/wiki/LOEventGenerationBias' *
 #*********************************************************************
  %(bias_module)s = bias_module  ! Bias type of bias, [None, ptj_bias, -custom_folder-]
  %(bias_parameters)s = bias_parameters ! Specifies the parameters of the module.
+ %(custom_fcts)s = custom_fcts ! List of files containing user hook function
```
oliviermattelaer, line 87: pass bias modules as hidden and change dummy model to call a function in the dummy file

oliviermattelaer (commented):
I have auto-reviewed this, such that I can include this into a first alpha release of 3.5.0
The only point that I think would be important to allow the same functionality as the NLO bias function within the custom_fcts format.
I would keep that old feature as hidden (not appearing in the run_card) but will extend the feature of custom_fcts to allow new class of function if they start by "user_" this should allow to cover all the use case of the bias module of valentin and allow a nice new feature as well.
- [x] change the run_card
- [x] add a new LO function for doing bias
- [x] automatically  add to the dummy_fct file all functions that start by user_

oliviermattelaer:
If they are not comment, I will push this branch (a fast forward) to 3.5.0 and then do an alpha tarball release for 3.5.0

PR #28, opened 2023-01-10
Fixes for build errors in lo_vect_diagid 

valassi:
Hi @oliviermattelaer these are a few fixes for your MR https://github.com/mg5amcnlo/mg5amcnlo/pull/5

Without these fixes, I generated gg to ttgg and it failed to build
```
/cvmfs/sft.cern.ch/lcg/releases/gcc/11.2.0-ad950/x86_64-centos7/bin/gfortran -w -fPIC  -ffixed-line-length-132 -w -c auto_dsig1.f -I../../Source/ -fopenmp
auto_dsig1.f:453:7:

  453 |       C.     COL_RAND: A LIST OF RANDOM NUMBER (WILL BE USED TO PICK
      |       1
Error: Unclassifiable statement at (1)
auto_dsig1.f:155:13:

  155 |       CHANNEL = SUBDIAG(1)
      |             1
Error: Symbol ‘channel’ at (1) has no IMPLICIT type
auto_dsig1.f:388:62:

  388 |       CALL SMATRIX1_MULTI(P_MULTI, HEL_RAND, COL_RAND, CHANNEL,
      |                                                              1
Error: Symbol ‘channel’ at (1) has no IMPLICIT type
``` 
With these fixes, the build succeeds.
Can you please check if they make sense?

That said, this patch fixes the build errors, but there are other issues in lo_vect_diagid
- the main reason why I wanted lo_vect_diagid was to check whether coloramps.inc and coloramps.h produces the same order: even with lo_vect_diagid, the order of colors remains different for gg->ttgg and gg->ttggg (but they are the same for tt and ttg)
- in addition, lo_vect_diagid changes the "AMP2(i)=AMP2(i)+AMP(j)*DCONJG(AMP(j))" lines, and I am now unable to get the same xsection in fortran and cudacpp (this is true whether I use coloramps.h as-is or I force it to hav ethe same order as coloramps.inc... but anyway I think that this file only affects the LHE file, not the xsec)

So I guess there is more work needed before #5 can be merged? However thes three patches I propose (or a variant thereof) are probably needed

Thanks!
Andrea

### DIFF
--- a/madgraph/iolibs/template_files/auto_dsig_v4.inc
+++ b/madgraph/iolibs/template_files/auto_dsig_v4.inc
@@ -89,6 +89,7 @@ c
 c     local
 c
 	   double precision P1(0:3, nexternal)
+      INTEGER CHANNEL
 C  
 C DATA
 C  
@@ -296,7 +297,7 @@ c     local
 c
 	   double precision P1(0:3, nexternal)
 	   integer ivec
-
+      INTEGER CHANNEL
 C  
 C DATA
 C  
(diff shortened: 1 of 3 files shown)


valassi:
This was related to lo_vect_diag https://github.com/mg5amcnlo/mg5amcnlo/pull/5, but we agreed with Olivier to not merge lo_vect_diag. I am therefore closing also this MR #28 as WONTMERGE.

PR #30, opened 2023-01-25
first version to show andrea

oliviermattelaer:
Hi Andrea,

as a follow up on the previous merge,
I'm starting here to setup the code with the normal user-interface 
such that we can have
1) batch sixe (used)
2) vec size (my locksize)

To be sure that we are on the same level of understanding


oliviermattelaer:
Note for myself:
1) need to update SUBDIAG AND IB to a size (NB_BLOCK -> currently set to VECSIZE_MEMMAX which is too large but ok)
2) need to update IPSEL from scalare to grid size (only done in auto_dsigX.f)
3) need to upgrade channel to a list (size NB_BLOCK) (this is work in progress) -> need to change the bridge!
4) probably need to change the signature of DSIGX_VEC (not really needed but can be more clear how channel is passed)

oliviermattelaer:
That branch has been abandonned and the idea restarted with another branch name (with warp) -> closing

PR #35, opened 2023-02-08
Allow use of python's multiprocessing to concurrently calculate systematics

Dr15Jones:
Motivation:
When monitoring a 2200 event NLO generation using multiple cores for the CMS experiment, we noticed that approximately 300 seconds were being used to calculate the systematics. After this change, the same generation only took 45 seconds.

Changes:
In order to make it easier to reason about what was actually changing during the call to `Systematics.run`, I refactored the code to isolate those components which could safely be used via multiprocessing. This change also made it easier to create unit tests for the different components.

The `run` step now takes an Executor to decide how to actually do the processing. The SerialExecutor runs all work within the original python process. The ParallelExecutor reads the event in the original process and passes the event to N different processes (where N is the number of cores to use). These processes update the event and pass the event to a final process. The final process is the one which writes the events to a file (this guarantees serial access to the output file). Once all events have been read, the original process waits for all the other processes to finish.

Testing:
The new unit test pass. Unfortunately, running `./tests/test_manager.py -p A` on the main branch (i.e. without these changes) leads to many failures. However, after applying these changes, that call gives back identical successes/failures.

A version of this code was used with MadGraph 2.9.9 within the CMS experiment's generation process and identical events were generated when comparing the original code to the changes equivalent to this pull request.

### DIFF
--- a/madgraph/interface/common_run_interface.py
+++ b/madgraph/interface/common_run_interface.py
@@ -1863,8 +1863,14 @@ class CommonRunCmd(HelpToCmd, CheckValidForCmd, cmd.Cmd):
             logger.warning('impossible to download all the pdfsets. Bypass systematics')
             return
         
-        if self.options['run_mode'] ==2 and self.options['nb_core'] != 1:
-            nb_submit = min(int(self.options['nb_core']), nb_event//2500)
+        cores_to_use = self.options['nb_core']
+        if self.options['run_mode'] ==2 and cores_to_use != 1:
+            nb_submit = min(int(cores_to_use), nb_event//2500)
+            #4 cores are about as fast as using 3 clusters
+            if 4*nb_submit < 3*cores_to_use:
+                nb_submit = 1
+            else:
+                cores_to_use = 1
         elif self.options['run_mode'] ==1:
             try:
                 nb_submit = min(int(self.options['cluster_size']), nb_event//25000)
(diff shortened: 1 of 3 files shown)


Dr15Jones:
@makortel FYI

oliviermattelaer:
This sounds really cool and impressive.

They are issue to be fixed for sure since this seems based on 2.9.9 which is at bug fix level (so speed improvment are normally excluded from such release) and that you seem to miss feature from 3.4.0 (but I guess that this should not be too problematic but we need for sure a tecnical merge in a LTS branch before porting this to the 3.5.0 version.

Now I'm thinking to create a 2.10.0 release with some "IO and speed improvment" for CMS, so that would solve the merging in a simpler way. But I need to contact CMS to see how they react to that.

Dr15Jones:
Hmm, I thought I did the pull request into the 'main' branch since that is the branch upon which I rebased (and then fixed) all the changes. Don't know why this went to LTS_RO_NOLgridpack instead.

Dr15Jones:
Ah, I see now that you changed the target branch.

Dr15Jones:
> Now I'm thinking to create a 2.10.0 release with some "IO and speed improvment" for CMS, so that would solve the merging in a simpler way.

I actually have a 2.9.11 based branch with the change 

https://github.com/Dr15Jones/mg5amcnlo/tree/parallel_systematics_2_9_11

oliviermattelaer:
Yes because this the branch that might become the 2.10.0.
But that creates more issue that I was hoping to resolve so I will change it to something else for clarity.
And then we will see where/how to merge this

oliviermattelaer:
Ok I have create a pull request for that 2.9.11 base branch:
https://github.com/mg5amcnlo/mg5amcnlo/pull/37

Thanks,

olivier

PR #37, opened 2023-02-08
Allow use of python's multiprocessing to concurrently calculate systematics (from 2.9.11)

oliviermattelaer:
This should make it easier to merge with 2.9.x if needed


oliviermattelaer:
@Dr15Jones such that you are pinged here as well

PR #40, opened 2023-03-01
chore: Remove Python 2 only code blocks as Python 2 support dropped

matthewfeickert:
Given https://github.com/mg5amcnlo/mg5amcnlo/commit/25a208db24bf9c28f6650f10b86fdfa04a7b1493 all Python 2 support has been dropped, and so code that references or uses Python 2 can be removed. This should also extend to removing all use of `six`, as it is used only for Python 2 to Python 3 support.

Note that **all code under `vendor/`** has been left alone as it is assumed that this is literally ported code and for ease of updating in the future it should be identical to original source code.

To simplify comparisons to the Python version use the pattern of

```python
sys.version_info < (3, 7)
```

as comparison to tuples is supported.

File specific comments:

  - `aloha/bin/aloha`
  - `bin/create_aloha_release.py`
  - `Template/MadWeight/bin/madweight.py`
  - `Template/MadWeight/bin/mw_options`
  - `tests/test_manager.py`
    
    All had `sys.exit` conditions that claimed their code didn't work with Python 3. This is assumed to now be false, as otherwise these pieces of code would no longer work in modern `MadGraph5_aMC@NLO`, and so these `sys.exit` conditions are removed.

  - `bin/mg5_aMC`

    Use standard 'python -m pip' pattern to avoid potential issues with the wrong version of `pip3` being on PATH.

  - `madgraph/iolibs/export_v4.py`
  - `madgraph/various/misc.py`

    In Python 3.x `f2py` has the following names under the virtual environment's `bin/` dir: `f2py`, `f2py3`, `f2py3.x`. As the patterns that exist are for `'f2py2.x'` and `'f2py-2.x'` and these patterns only exist for Python 2 versions of NumPy the relevant code blocks are removed.

* Remove use of `from __future__ import print_function`.

### DIFF
--- a/madgraph/interface/madgraph_interface.py
+++ b/madgraph/interface/madgraph_interface.py
@@ -18,7 +18,6 @@
 from __future__ import division
 
 from __future__ import absolute_import
-from __future__ import print_function
 import atexit
 import collections
 import cmath
--- a/madgraph/interface/madgraph_interface.py
+++ b/madgraph/interface/madgraph_interface.py
@@ -1547,17 +1547,6 @@ This will take effect only in a NEW terminal
             except Exception:
                 raise self.InvalidCmd('%s needs argument True or False'%args[0])
 
-        if args[0] in ['low_mem_multicore_nlo_generation']:
-            if args[1]:
-                if sys.version_info[0] == 2:
-                    if  sys.version_info[1] == 6:
-                        raise Exception('python2.6 does not support such functionalities please use python2.7')
-                #else:
-                #    raise Exception('python3.x does not support such functionalities please use python2.7')
-        
-
-
-
         if args[0] in ['gauge']:
             if args[1] not in ['unitary','Feynman', 'axial']:
                 raise self.InvalidCmd('gauge needs argument unitary, axial or Feynman.')
--- a/madgraph/various/process_checks.py
+++ b/madgraph/various/process_checks.py
@@ -1267,12 +1267,9 @@ class LoopMatrixElementTimer(LoopMatrixElementEvaluator):
         dir_name=res_timings['dir_path']
 
         def check_disk_usage(path):
-            return subprocess.Popen("du -shc -L "+str(path), \
-                stdout=subprocess.PIPE, shell=True).communicate()[0].decode(errors='ignore').split()[-2]
-            # The above is compatible with python 2.6, not the neater version below
-            # -> need to check if need .decode for python3.7
-            #return subprocess.check_output(["du -shc %s"%path],shell=True).\
-            #                                                         split()[-2]
+            return subprocess.check_output(
+                [f"du -shc {path}"], shell=True, encoding="utf8"
+            ).split()[-2]
 
         res_timings['du_source']=check_disk_usage(pjoin(\
                                                  export_dir,'Source','*','*.f'))
--- a/INSTALL
+++ b/INSTALL
@@ -24,7 +24,7 @@ Dependencies :
 
  * MadGraph5_aMC@NLO *
 
-- Python 3.7 (or higher, python2.7 is also working for the time being)
+- Python 3.7 (or higher)
 
  * MadEvent *
 Package for the LO cross-section computation and generation of events 
--- a/MadSpin/decay.py
+++ b/MadSpin/decay.py
@@ -2,7 +2,6 @@
 
 from __future__ import division
 from __future__ import absolute_import
-from __future__ import print_function
 from madgraph.interface import reweight_interface
 from six.moves import map
 from six.moves import range
--- a/MadSpin/interface_madspin.py
+++ b/MadSpin/interface_madspin.py
@@ -15,7 +15,6 @@
 """ Command interface for MadSpin """
 from __future__ import division
 from __future__ import absolute_import
-from __future__ import print_function
 import collections
 import logging
 import math
--- a/MadSpin/madspin
+++ b/MadSpin/madspin
@@ -20,9 +20,9 @@ and call immediately the command line interface scripts"""
 from __future__ import absolute_import
 from __future__ import print_function
 import sys
-if  sys.version_info[1] < 6:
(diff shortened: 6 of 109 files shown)


oliviermattelaer:
Hi,

Thanks a lot for this. 
They are just one change (on the help/comment of the convert model that I would not include).

Thanks so much,

Olivier

madgraph/interface/madgraph_interface.py
```diff
@@ -674,11 +674,11 @@ def help_add(self):
         logger.info("   --output=  : Specify the name of the directory where the merge is done.")
         logger.info("                This allow to do \"import NAME\" to load that merge.")
         logger.info("   --recreate : Force to recreated the merge model even if the merge model directory already exists.")
-        
+
     def help_convert(self):
         logger.info("syntax: convert model FULLPATH")
-        logger.info("modify (in place) the UFO model to make it compatible with both python2 and python3")
-        
+        logger.info("modify (in place) the UFO model to make it compatible with python3")
```
matthewfeickert, line 680: > They are just one change (on the help/comment of the convert model that I would not include).

@oliviermattelaer you're just saying that this change should be reverted, correct? 

If so, I'll get this tomorrow and then also clean up the commits a bit and squash them into 1 to make it cleaner.

It looks like the removal of all use of `six` is probably going to be big enough that I should do that in another PR that only touches that.

Beyond using a Docker container that has all the other dependencies setup in it, is there a way to run the tests suite

https://github.com/mg5amcnlo/mg5amcnlo/blob/57eea3e92d06d2fb27cd3c0e66ced15ed2d05a44/README.developer#L33-L43

automatically?

madgraph/various/process_checks.py
```diff
@@ -1267,12 +1267,8 @@ def time_matrix_element(self, matrix_element, reusing = False,
         dir_name=res_timings['dir_path']
 
         def check_disk_usage(path):
-            return subprocess.Popen("du -shc -L "+str(path), \
-                stdout=subprocess.PIPE, shell=True).communicate()[0].decode(errors='ignore').split()[-2]
-            # The above is compatible with python 2.6, not the neater version below
-            # -> need to check if need .decode for python3.7
-            #return subprocess.check_output(["du -shc %s"%path],shell=True).\
-            #                                                         split()[-2]
+            return subprocess.check_output([f"du -shc {path}"], shell=True).\
```
matthewfeickert, line 1270: Use python f-strings here as they were added in Python 3.6 and this is Python 3.7+ code.

matthewfeickert:
@oliviermattelaer this is ready for review. Is there a standard test runner environment that you ask people to use (c.f. https://github.com/mg5amcnlo/mg5amcnlo/pull/40#discussion_r1121411410) before requesting a review normally?

madgraph/various/process_checks.py
matthewfeickert: RE: the comment

```
# -> need to check if need .decode for python3.7
```

this can be easily dealt with by using the standard `encoding="utf8"` to get strings back instead of bytestrings.

Example:

```console
$ docker run --rm -ti python:3.10 /bin/bash
root@457ba718d9b0:/# cd home/
root@457ba718d9b0:/home# du -shc $PWD
4.0K	/home
4.0K	total
root@457ba718d9b0:/home# python
Python 3.10.10 (main, Feb 11 2023, 02:40:33) [GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from pathlib import Path
>>> import subprocess
>>> path = Path.cwd()
>>> subprocess.check_output([f"du -shc {path}"], shell=True).split()[-2]
b'4.0K'
>>> subprocess.check_output([f"du -shc {path}"], shell=True, encoding="utf8").split()[-2]
'4.0K'
>>>
```

matthewfeickert:
@oliviermattelaer Should this PR be targeting the `LTS_dev` branch?

oliviermattelaer:
No for the  LTS, we pledge to be super conservative in term of update.
With only minimal bug fixing. The idea is that any change to the LTS should be usable by CMS/Atlas without doing any phyiscs validation. 
While we officially do not support python2.7 within the LTS and do not test for that python version anymore.
We are still allowing the user to run with python2.7 (with warning). So, I will not drop on purpose the compatibility with python2.7 for that version of the code.

Olivier

PS: I'm back in Europe and will do the merge (likely not in the main directly) on Tuesday.

matthewfeickert:
> No for the LTS, we pledge to be super conservative in term of update. With only minimal bug fixing. The idea is that any change to the LTS should be usable by CMS/Atlas without doing any phyiscs validation. While we officially do not support python2.7 within the LTS and do not test for that python version anymore. We are still allowing the user to run with python2.7 (with warning). So, I will not drop on purpose the compatibility with python2.7 for that version of the code.

Okay, thanks, this is useful. So if I understand correctly then the default branch to target for contributions is `main`, correct?

> PS: I'm back in Europe and will do the merge (likely not in the main directly) on Tuesday.

Sounds good. No rush. :+1:

oliviermattelaer:
Not sure I fully understood those github pull request here.
But looks like I have not pushed your commit within this one:
https://github.com/mg5amcnlo/mg5amcnlo/pull/41

So I will close this thread.

Thanks a lot for this.

matthewfeickert:
> Not sure I fully understood those github pull request here.

@oliviermattelaer You didn't need to close this PR and make PR #41. You can simply change the target branch, which is why I was asking you what branch PRs should be targetting in https://github.com/mg5amcnlo/mg5amcnlo/pull/40#issuecomment-1455430872.

Apparently the answer is all new contributions should target branch `3.5.0`? This really isn't clear at all especially as the default branch for the repository is `main`. Can you please add some developer docs that explain the workflow and how to test contributions before requesting review?

oliviermattelaer:
Yeah sorry about that, I've not enough experience in this website. launchpad was simpler to handle.
Actually I tried what you said but end up in this mess. Not sure why.

PR #41, opened 2023-03-08
Feat/drop python2

oliviermattelaer:
Content from matthew to clean reference from python2


matthewfeickert:
Future people should refer to PR #40 for information.

PR #43, opened 2023-03-09
chore: Apply pyupgrade --py37-plus

matthewfeickert:
Apply the effects of `pyupgrade --py37-plus` to the codebase by running:

```
find . -type f -iname "*.py" | grep -v "vendor/" | xargs pyupgrade --py37-plus
```

This applies multiple upgrade steps but the main ones are:

- Remove unneeded 'from __future__' imports
- Remove trivial use of six
- Use f-strings whenever possible
- Use OSError over IOError
- Drop unneeded inheritance from object
- Use simplified super()

* Fix incorrect indenting in tests/test_manager.py for bypass_for_py3. This should be removed in a follow up PR.

Note that this commit is _huge_ because is is applying automated code changes across the code base. There are some things that this PR misses, but it gets most of them.

Examples of things missed that should be fixed in the future:
* Trivial imports of `six` when `six` is no longer used.
* Imports of `six.moves` and other `six` modules that are no longer needed.
* The use o the `bypass_for_py3` decorator as it is no longer needed.

**Note:** `pyupgrade` assumes that it is starting with **valid Python 3 code**. There may be some instances where code here was really only Python 2 code and hadn't been properly updated, which means that some things might not have gotten transformed correctly. **This PR needs to be tested before merging.**

What is the proper dev environment for running tests?


oliviermattelaer:
So I did try 
pyupgrade --py36-plus --keep-percent-format

But they are a couple of the tests that fails to pass:
The most critical issue is that it overwrites a lot of "set" to python keyword.
But since we allow to over-ride the meaning of "set" to have it optionally ordered to ensure reproducibility.

```
if madgraph.ordering:
    set = misc.OrderSet
```

So we will likely need to hack the code to remove all the rules related to "set" to make it possible.

matthewfeickert:
@oliviermattelaer I'm not really sure why you closed this given what you posted. Also I'm not sure what you're even talking about in your repsonse. You can clearly see that there were no keyword overrides in the PR, so I'm not sure what you mean.

> But since we allow to over-ride the meaning of "set" to have it optionally ordered to ensure reproducibility.

Why do you override a builtin instead of just creating your own ordered set?

oliviermattelaer:
One of the potential issue is the change of the following lines:
```
-dirs = set([j['dir'] for j in jobs])
+dirs = {j['dir'] for j in jobs}
``` 
Where the "{" syntax does not behaves like the set command above.

This makes some CI to crash due to operation between the two type of set which is not supported.
And if I do implement them, they are still other CI failing by returning number which are not the expected ones.
(in particular in loop computation).

> Why do you override a builtin instead of just creating your own ordered set?

That was the easiest solution when migrating from python2 to python3 while keeping the ordered set of python2.
The fully unordered feature of python creates a lot of trouble for code validation since generated code is not reproducible when using the set in python3. I know set in python2 were also unordered but this was much less problematic in python2 since it was reproducible per machine (which was enough for the CI in most cases)).

PR #46, opened 2023-03-29
Fixed typo in UFO parser for cpp output

zeniheisser:
Fixed small bug where non-standard UFO models could not be exported using cpp output (simple typo)

### DIFF
--- a/madgraph/iolibs/ufo_expression_parsers.py
+++ b/madgraph/iolibs/ufo_expression_parsers.py
@@ -873,7 +873,7 @@ class UFOExpressionParserCPP(UFOExpressionParser):
         elif p[1] == 'reglog': p[0] = 'reglog' + p[2]
         elif p[1] == 'reglogp': p[0] = 'reglogp' + p[2]
         elif p[1] == 'reglogm': p[0] = 'reglogm' + p[2]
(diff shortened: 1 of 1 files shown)


zeniheisser:
Tagging @valassi. Small typo in the UFO parser means output to cpp generally will crash after switching model.

valassi:
Hi @zeniheisser thanks. Can you transform this into a PR against the gpucpp branch please? (@oliviermattelaer do you agree? I think it is better than main, so we can directly use it in our GPU developpments)

zeniheisser:
@valassi I think it fits here, since it is an upstream bug, but we certainly should fix it in the gpucpp branch as well

valassi:
Thanks again Zenny. I committed this to gpucpp (I wanted to make a MR but I made a mistake, anyway it is a really minor fix)
https://github.com/mg5amcnlo/mg5amcnlo/commit/8512c16a49dc96adb21505b9359ba811b0d169ca

oliviermattelaer:
Merged in https://github.com/mg5amcnlo/mg5amcnlo/commit/550172de74a297af7d175bf8a3c458c170b0792d
(in LTS_dev)

PR #31, opened 2023-01-25
Minor 'vecsize3' changes to address Olivier's comments to the 'vecsize2' MR

valassi:
Hi @oliviermattelaer these are some minor changes to address your comments to https://github.com/mg5amcnlo/mg5amcnlo/pull/24

Is it better like this?

Thanks
Andrea

PR #49, opened 2023-04-11
Fixes in vecsize (define VECSIZE_MEMMAX_COUPL in addition to VECSIZE_MEMMAX)

valassi:
Hi @oliviermattelaer @roiser this is the full set of upstream fixes for https://github.com/madgraph5/madgraph4gpu/issues/629

My previous two MRs vecsize (https://github.com/mg5amcnlo/mg5amcnlo/pull/23) and vecsize2 (https://github.com/mg5amcnlo/mg5amcnlo/pull/24) had broken the "launch" functionality for a simple Fortran case. Now this is fixed.

The fix is ugly, as it introduces two separate variables VECSIZE_MEMMAX_COUPL and VECSIZE_MEMMAX (and a sanity check that they are equal). But it's as good as it gets (or at least as good as I was able to make it), given that the Fortran include statement does not use the preprocessor include guards, and that I did not want to modify every Fortran "include" in madgraph into a "#include". Anyway the values are set by code generation anyway.

I also include the minor fixes that were previosuly in https://github.com/mg5amcnlo/mg5amcnlo/pull/31 (which I will close), addressing some changes suggested by Olivier.

All in all, the code now looks much more like the madgraph previous to my two changes - which is good IMO.

@oliviermattelaer I set you as reviewer - but I will base all my developments in madgraph4gpu on this, so if you can have a quick look that would be great. Otherwise I can merge it (on gpucpp, anyway), and we discuss possible improvements later on.

Thanks Andrea

### DIFF
--- a/Template/LO/SubProcesses/reweight.f
+++ b/Template/LO/SubProcesses/reweight.f
@@ -1790,7 +1790,7 @@ C
       include 'genps.inc'
       include 'run.inc'
       include 'nexternal.inc'
-      include 'vector.inc' ! needed by coupl.inc (defines VECSIZE_MEMMAX)
+      include 'vector.inc'
       include 'coupl.inc'
 C      include 'maxparticles.inc'
       
@@ -1817,7 +1817,7 @@ C
       include 'genps.inc'
       include 'run.inc'
       include 'nexternal.inc'
-      include 'vector.inc' ! needed by coupl.inc (defines VECSIZE_MEMMAX)
+      include 'vector.inc'
       include 'coupl.inc'
 C      include 'maxparticles.inc'
       
--- a/Template/LO/SubProcesses/reweight.f
+++ b/Template/LO/SubProcesses/reweight.f
@@ -1790,7 +1790,7 @@ C
       include 'genps.inc'
       include 'run.inc'
       include 'nexternal.inc'
-      include 'vector.inc'
+      include 'vector.inc' ! needed by coupl.inc (defines VECSIZE_MEMMAX)
       include 'coupl.inc'
 C      include 'maxparticles.inc'
       
@@ -1817,7 +1817,7 @@ C
       include 'genps.inc'
       include 'run.inc'
       include 'nexternal.inc'
-      include 'vector.inc'
+      include 'vector.inc' ! needed by coupl.inc (defines VECSIZE_MEMMAX)
       include 'coupl.inc'
 C      include 'maxparticles.inc'
       
--- a/Template/LO/SubProcesses/reweight.f
+++ b/Template/LO/SubProcesses/reweight.f
@@ -1790,7 +1790,7 @@ C
       include 'genps.inc'
       include 'run.inc'
       include 'nexternal.inc'
-c     include 'vector.inc' ! needed by coupl.inc (defines VECSIZE_MEMMAX)
+c     include 'vector.inc' ! defines VECSIZE_MEMMAX
       include 'maxamps.inc'
       include 'cluster.inc'
       include 'coupl.inc'
@@ -1819,7 +1819,7 @@ C
       include 'genps.inc'
       include 'run.inc'
       include 'nexternal.inc'
-c     include 'vector.inc' ! needed by coupl.inc (defines VECSIZE_MEMMAX)
+c     include 'vector.inc' ! defines VECSIZE_MEMMAX
       include 'maxamps.inc'
       include 'cluster.inc'
       include 'coupl.inc'
--- a/madgraph/iolibs/export_v4.py
+++ b/madgraph/iolibs/export_v4.py
@@ -6941,10 +6941,8 @@ class UFO_model_to_mg4(object):
         file = open(os.path.join(MG5DIR,\
                               'models/template_files/fortran/rw_para.f')).read()
 
-        includes=["include \'../vector.inc\'",
-                  "include \'coupl.inc\'",
-                  "include \'input.inc\'",
-                  "include \'model_functions.inc\'"]
+        includes=["include \'coupl.inc\'","include \'input.inc\'",
+                                              "include \'model_functions.inc\'"]
         if self.opt['mp']:
             includes.extend(["include \'mp_coupl.inc\'","include \'mp_input.inc\'"])
         # In standalone and madloop we do no use the compiled param card but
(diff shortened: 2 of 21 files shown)


Template/LO/SubProcesses/reweight.f
```diff
@@ -1790,7 +1790,7 @@ subroutine update_scale_coupling(p, wgt)
       include 'genps.inc'
       include 'run.inc'
       include 'nexternal.inc'
-      include 'vector.inc' ! needed by coupl.inc (defines VECSIZE_MEMMAX)
+      include 'vector.inc'
```
oliviermattelaer: Why not keeping the comment here (would be more clear why this is included)

Template/LO/SubProcesses/reweight.f
```diff
@@ -1817,7 +1817,7 @@ subroutine update_scale_coupling_vec(all_p, all_wgt,all_q2fact, VECSIZE_USED)
...
```
oliviermattelaer: same here

madgraph/iolibs/export_v4.py
```diff
@@ -6941,10 +6949,8 @@ def copy_standard_file(self):
         file = open(os.path.join(MG5DIR,\
                               'models/template_files/fortran/rw_para.f')).read()
 
-        includes=["include \'../vector.inc\'",
-                  "include \'coupl.inc\'",
-                  "include \'input.inc\'",
-                  "include \'model_functions.inc\'"]
+        includes=["include \'coupl.inc\'","include \'input.inc\'",
```
oliviermattelaer, line 6952: Here I'm confused why vector.inc is not include here. (or why it is needed above)

madgraph/iolibs/export_v4.py
```diff
@@ -7027,10 +7033,17 @@ def create_coupl_inc(self):
...
         header = """C
-C NB: VECSIZE_MEMMAX is defined in vector.inc
-C NB: vector.inc must be included before coupl.inc
+C     NB: VECSIZE_MEMMAX is defined in vector.inc, but this is not
+C     included in coupl.inc because Fortran has no include guards
+C
+C     NB: therefore coupl.inc introduces the definition of a separate
+C     variable VECSIZE_MEMMAX_COUPL, whose value must be the same as
+C     that of VECSIZE_MEMMAX (this is cross-checked in PROGRAM DRIVER)
 C
```
oliviermattelaer, line 7042: Ok, I see now, i would prefer the previous option is to use 
include 'vector.inc' before each line include 'coupl.inc'
with a comment on why the include to vector.inc is needed

oliviermattelaer (changes requested):

madgraph/iolibs/export_v4.py
valassi: Hi Olivier (cc @roiser) thanks for the feedback here.

> Ok, I see now, i would prefer the previous option is to use include 'vector.inc' before each line include 'coupl.inc' with a comment on why the include to vector.inc is needed

I am afraid this is just too complex. If I understand correctly, you suggest that I try again along the lines of my "old" patches, i,e. include vector.inc before each and every coupl.inc. The problem is that doing this (or at least, doing it the way I tried to do it) eventually results in problems elsewhere, see issue #629. Sometimes I get VECSIZE_MEMMAX defined twice, sometimes I get it not defined at all. This is why I suggest here to go back closer to your initial implementation of last year, and use two different variables with the same value, VECSIZE_MEMMAX and VECSIZE_MEMMAX_COUPL.

In your original implementation last year these two variables were quite simply hardcoded, eg 16384 if this is the value specified in the output vecsize parameter. I still prefer having variables, rather than hardcoding, this is maybe where you disagree and prefer whatever comes from code generation. I prefer variables because you can change them easily from the code in one place, without redoing the full generation.

Maybe this is what the runcard should be doing? But I have not yet integrated the runcard in the rest of the makefile machinery.

I suggest the following option 1:
- for the moment, I keep this patch as-is with minor improvements (I need to rebase it anyway over other changes in the last month): in particular I keep the idea of having two variables with different names and the same value, but I do add back a few more comments about the fact that VECSIZE_MEMMAX is defined in vector.inc
- later on, we discuss how to improve this using runcards

The option 2 that you were proposing above does not work, it is too complex: if I try again to have vector.inc before coupl.inc whenever needed and only when needed, this will give me #629, and I was unable to fix that in a reasonable time.

The only other alternative I see is option 3: go back from VECSIZE_MEMMAX and VECSIZE_MEMMAX_COUPL to 16384 hardcoded. I would really prefer to avoid this.

If the Fortran include directive was less nonsensical (i.e. if we could use include guards and the preprocessor on every include), this would have been solved in 10 seconds. Since Fortran include is what it is, the only other far fetched alternative is option 4 where we use #include instead of include EVERYWHERE, but this I would also avoid. It would mean touching too much code, and would certainly go wrong. 

Are you ok with option 1? Thanks
Andrea

valassi:
Hi @oliviermattelaer (cc @roiser) I take this back.

Olivier, you suggested what I call option 2,
> Ok, I see now, i would prefer the previous option is to use include 'vector.inc' before each line include 'coupl.inc' with a comment on why the include to vector.inc is needed

I argued above that this is complicated, and I was unable to fix https://github.com/madgraph5/madgraph4gpu/issues/629. It turns out that (not surprisingly!) you were able to fix it very fast, with your commit
https://github.com/mg5amcnlo/mg5amcnlo/commit/c1d10f42631378d4c5ea6f594c9318cedff72953

I will therefore keep your fix for that issue, and my PR here becomes mostly unnecessary. (I will strip it donw to include only some partis of it in a new one, maybe).

So I am closing this. Thanks for the help, sorry for the noise.

valassi:
This is now superseded by the simpler #54.

Template/LO/SubProcesses/reweight.f
valassi: done in #54

Template/LO/SubProcesses/reweight.f
valassi: done in #54

madgraph/iolibs/export_v4.py
valassi: This one I specifically did **NOT** fix in #54 yet. Indeed something would be needed there. I suggest to look at it as issues show up.

PR #50, opened 2023-04-12
fix the to_pdf common block bug

huashengshao:
Hi, I just quickly fixed the problem with the to_pdf common block, which now uses the include '/PATH/TO/pdf.inc' instead of hard-coded the common block.


marcozaro:
Confirmed.
The relevant tests pass.

./tests/test_manager.py -t0 -p A test_generate_eett_nlo_emela_bs test_generate_eett_lo_emela_clusterpath test_generate_eett_nlo_qcd_noisr test_generate_eeww_nlo_emela_wph test_generate_eeww_nlo_emela_noph

PR #54, opened 2023-05-17
Minor 'vecsize3' changes (improve passcuts interface and add comments about vector.inc)

valassi:
This is a (final?) followup to my previous MRs vecsize (https://github.com/mg5amcnlo/mg5amcnlo/pull/23) and vecsize2 (https://github.com/mg5amcnlo/mg5amcnlo/pull/24). 
- First, it includes a few improvements suggested by @oliviermattelaer in #24 about the passcuts interface, which now includes VECSIZE_USED (https://github.com/mg5amcnlo/mg5amcnlo/pull/24#discussion_r1087085378). This patch was originally in MR #31, which had been closed without merging.
- Second, it consistently adds comments about VECSIZE_MEMMAX whenever vector.inc, coupl.inc or cluster.inc are included, as discussed with @oliviermattelaer in #49.

This MR replaces #49, which has been closed. The main point of that MR was to fix https://github.com/madgraph5/madgraph4gpu/issues/629, which I had attempted to do by adding VECSIZE_MEMMAX_COUPL to the coupl.inc file. However the actual issue ("launch" was broken because symmetry.f was not building) was instead fixed by Olivier in https://github.com/mg5amcnlo/mg5amcnlo/commit/c1d10f42631378d4c5ea6f594c9318cedff72953. It is then unnecessary and cumbersome to add a different VECSIZE_MEMMAX_COUPL and Olivier suggested to keep things as they were before (what I call "option 2" in #49).

**NB** : I imagine that there are other issues hidden in the code because of this choice. For instance, export_v4.py creates code which includes coupl.inc without including vector.inc first. I propose that we will try to fix these issues as they come along. *For the moment I have not attempted to fix export_v4.py*.

cc @roiser

### DIFF
--- a/Template/LO/Source/dsample.f
+++ b/Template/LO/Source/dsample.f
@@ -17,7 +17,7 @@ c                    but the SIMD vector size and GPU warp size are smaller)
 c**************************************************************************
       implicit none
       include 'genps.inc'
-      include 'vector.inc'
+      include 'vector.inc' ! defines VECSIZE_MEMMAX
 c     
 c Arguments
 c
@@ -667,7 +667,7 @@ c
       include 'genps.inc'
       include 'maxconfigs.inc'
       include 'run.inc'
-      include 'vector.inc'
+      include 'vector.inc' ! defines VECSIZE_MEMMAX
 c
 c     Arguments
 c
--- a/Template/LO/Source/dsample.f
+++ b/Template/LO/Source/dsample.f
@@ -12,6 +12,8 @@ c     dsig           Function to be integrated
 c     ninvar         Number of invarients to keep grids on (s,t,u, s',t' etc)
 c     nconfigs       Number of different pole configurations 
 c     VECSIZE_USED   Number of events in parallel out of VECSIZE_MEMMAX
+c                    (NB this is the #events handled by the cudacpp bridge,
+c                    but the SIMD vector size and GPU warp size are smaller)
 c**************************************************************************
       implicit none
       include 'genps.inc'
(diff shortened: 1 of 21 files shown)


oliviermattelaer:
Ok let me run a couple of sanity check on my side and see that I can (still) run the example that were working before and potentially fix associated issue and then I will merge this (let me try by the end of the week).

Thanks a lot for this,

olivier

valassi:
Thanks a lot to you Olivier :-)

Indeed I agree that removing th eugly VECSIZE_MEMMAX_COUPL is better, I am happy this is working ok. Again, I am *sure* that there are still issues to be fixed, but I suggest to first merge this, and then fix those later. If you look at export_v4.py, those coupl.inc without vector.inc, they are bound to fail: maybe you can create som eprocess where this code is activated, and test that it gets fixed if you make some fixes.

I am relying in this MR in https://github.com/madgraph5/madgraph4gpu/pull/654 in the GPU plugin. After this, we are a bit more in sync again...

oliviermattelaer:
Perfect, my tests were going trough -> merged

valassi:
Hi @oliviermattelaer (cc @roiser) thanks for the merge, very good!

Very minor point for possible future improvements, I was thinking that one way to make the treatment of vector.inc even better may be to remove it ALSO from cluster.inc?
- Now we must include vector.inc before coupl.inc; BUT if we include both clusters.inc and coupl.inc then we muct include cluster.inc before coupl.inc, and not include vector.inc
- In the future otherwise we might include vector.inc before either of clusters.inc or coupl.inc
Really a minor point, but maybe eventually this may be more maintainable/understandble. Let me know...

More importantly: I am moving to this mg5amcnlo #54 in the gpu plugin https://github.com/madgraph5/madgraph4gpu/pull/654. However there is a problem in ggttgg lhe files, https://github.com/madgraph5/madgraph4gpu/issues/655. I suspect that this might come from one of your earlier changes?

Anyway, what I suggest is to go ahead and use this in the GPU plugin. Then if we need a fix in mg5amcnlo we do that and we again update the gPU plugin. This MR %4 can stay nicley merged and closed.

PR #56, opened 2023-05-19
NOOP - Temporarely revert and reapply Olivier's select_color changes

valassi:
Hi @oliviermattelaer @roiser this is a NOOP MR on the upstream mg5amcnlo, which contains one commit where Olivier's https://github.com/mg5amcnlo/mg5amcnlo/commit/0a1038d97d8fc9d0d02556a39d14540778aa080b is reverted, then applied again.

I need this because I want an upstream commit that I can use as reference for the gpu plugin as-is, without mixing yet another bunch of things. I will go ahead and merge it: in any case it is a noop overall. But I need a commit hash to use as reference for the GPU plugin.

This is related to https://github.com/madgraph5/madgraph4gpu/issues/655, which most likely we need to separately fix in the GPU plugin.

Thanks
Andrea

### DIFF
--- a/madgraph/iolibs/template_files/matrix_madevent_group_v4.inc
+++ b/madgraph/iolibs/template_files/matrix_madevent_group_v4.inc
@@ -245,7 +245,7 @@ c         Set right sign for ANS, based on sign of chosen helicity
     ENDIF
     ANS=ANS/DBLE(IDEN)
 
-    call select_color(rcol, jamp2, channel,%(proc_id)s,  icol)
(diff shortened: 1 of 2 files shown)


valassi:
Note, in the meantime I "permamnetly" reverted Olivier's change in https://github.com/mg5amcnlo/mg5amcnlo/commit/42cf68bf37d7c3ef240b47e0da45008a9dbecdb4

Otherwise I have all the failures described in https://github.com/madgraph5/madgraph4gpu/issues/655

PR #62, opened 2023-06-24
Shower card

oliviermattelaer:
This modifies the shower_card python representation to be at the same level as the run_card.
and include additional entry in the shower_card specific to pythia8.


oliviermattelaer:
Hi Richard, Paolo,

I guess this is basically ready to be merged or should be fully trashed.
Whatever solution is fine for me.
Obviously this can not be merged "right" now since at least one unittest is failing.
But it sounds that this just need to update the tests. I will wait anyway to check if they are other test failing.

Cheers,

Olivier

oliviermattelaer:
OK All tests are fine now.
So for me we can include this but if one of you complains about this change.

PR #65, opened 2023-08-22
WIP: fix standalone output for vector.inc production and usage

roiser:
This fix applies to the "gpucpp" branch of the repo. When running e.g. 

```
generate e+ e- > mu+ mu-
output standalone test --vector_size=32
```

there this will produce the error 

```
The compilation fails with the following output message:
	    cd MODEL; make
	    make[1]: Entering directory '/afs/cern.ch/work/r/roiser/sw/madgraph4gpu/MG5aMC/mg5amcnlo/guttu2.sa/Source/MODEL'
	    /opt/rh/gcc-toolset-11/root/usr/bin/gfortran -w -fPIC  -ffixed-line-length-132  -c -o couplings.o couplings.f
	    couplings.f:16: Error: Can't open included file '../vector.inc'
	    make[1]: *** [<builtin>: couplings.o] Error 1
	    make[1]: Leaving directory '/afs/cern.ch/work/r/roiser/sw/madgraph4gpu/MG5aMC/mg5amcnlo/guttu2.sa/Source/MODEL'
	    make: *** [makefile:59: ../lib/libmodel.a] Error 2
```

the reason being that the ```vector.inc``` file containing the vector size for the parallel execution is not being produced. This PR changes the relevant ```madgraph/iolibs/export_v4.py``` in three places

- it calls self.write_vector_size also for the SA output
- it moves the write_vector_size from the ```ProcessExporterFortranME``` class to its base ```ProcessExporterFortran``` so that it can be used also in ```ProcessExporterFortranSA```
- it adds one missing include of vector.inc in couplings.f which maybe was overlooked

### DIFF
--- a/madgraph/iolibs/export_v4.py
+++ b/madgraph/iolibs/export_v4.py
@@ -932,6 +932,53 @@ param_card.inc: ../Cards/param_card.dat\n\t../bin/madevent treatcards param\n'''
 
         return True
 
+    #===========================================================================
+    # write_vector_size
+    #===========================================================================
+    def write_vector_size(self, fsock):
+        """Write the vector.inc which indicates how many event are handle in parralel."""
+
+    
+        try:
+            vector_size = self.opt['output_options']['vector_size']
+        except KeyError:
+            vector_size = 1
+        vector_size = banner_mod.ConfigFile.format_variable(vector_size, int, name='vector_size')
+        vector_size = max(1, vector_size)
+        text=["""C
+C If VECSIZE_MEMMAX is greater than 1, a vector API is used:
+C this is designed for offloading MEs to GPUs or vectorized C++,
+C but it can also be used for computing MEs in Fortran.
+C If VECSIZE_MEMMAX equals 1, the old scalar API is used:
+C this can only be used for computing MEs in Fortran.
+C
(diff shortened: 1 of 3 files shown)


oliviermattelaer:
Thanks a lot Stefan,

Sorry to be nitpicking on this.
But I do not think that you need to move the function "write_vector_size" 
Can you revert it to his original place (this is just to avoid spurious conflict).

If you did try and this was not working then no problem. 
Consider this has an approval of merging and after this check (and revert if possible) you can merge yourself (if you can, otherwise will do it).

Thanks,

Olivier

oliviermattelaer:
hi @roiser,

What should we do about this PR?
I actually move the write_vector_size function recently (So  sorry you were right)
I do guess that the branch gpucpp_for360 actually contains most of this changes.
But I have never looked at what the code is expected to do for the following command:
```
generate e+ e- > mu+ mu-
output standalone test --vector_size=32
```
(which might have a model ready for 32 events, but no code using/needing such type of model)
Do you see a case for at that (at pure fortran level)?

roiser:
Hi Olivier, 

I just checked, it would still fail, but I agree the use case is now doubtful, it works though as expected with "standalone_<backend>", let me close this.

PR #68, opened 2023-10-04
allow set group_subprocesses gpu to work for madevent mode

oliviermattelaer:
just allow the following syntax to work:
```
set group_subprocesses gpu
generate p p > w+ j
output
```
which avoid to have to hack the export_v4.py file all the time.


roiser:
yeeesss, I wanted to ask you already about it but didn't dare to ;-) thanks lot

PR #77, opened 2024-02-24
Drop unnecessary line from oxxxxx.F

APN-Pucky:
It is already computed before the if. `ixxxxx.F` also does not recompute `pp`.

### DIFF
--- a/HELAS/oxxxxx.F
+++ b/HELAS/oxxxxx.F
@@ -81,7 +81,6 @@ c
             
          else
             
-            pp = min(p(0),dsqrt(p(1)**2+p(2)**2+p(3)**2))
             sf(1) = dble(1+nsf+(1-nsf)*nh)*rHalf
             sf(2) = dble(1+nsf-(1-nsf)*nh)*rHalf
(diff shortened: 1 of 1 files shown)


oliviermattelaer:
Thanks, for finding this out.
I will merge this right away.

However, I'm curious. This routine is only used for model using the Madgraph4 format (so before the community convention for UFO model).  Do you have a good reason to use such type of model? Because we would like to remove such support.

oliviermattelaer:
Maybe I should have wait,
But I checked the equivalent version for the UFO/CPP/Python and loop output
and they all have the same difference.  So I did change them all.
https://github.com/mg5amcnlo/mg5amcnlo/commit/702ddee6ed8632a2faaebbb5c986a812036273cc

APN-Pucky:
No particular reason/use-case from my side. I just used the madgraph HELAS as reference for a fresh-started fun-project of writing a gpu HELAS [pyhelas](https://gitlab.com/APN-Pucky/pyhelas) with [jax](https://github.com/google/jax) as gpu-driver/interface and stumbled over that line.

PR #79, opened 2024-02-28
3.6.0

marcozaro:


oliviermattelaer:
So two obvious issue here:
1) they are some conflict to fix between the branch
2) some of the unittest and acceptance test are currently crashing

PR #80, opened 2024-02-28
Pull request to merge the Asymmetric collisions version of the code to the main branch 

AntonSafr96:
Dear Oliver, Marco and Rikkert,

I wish to merge my version of the code with the main one, in order to make it a part of the official release. 
I'm not sure about the proper procedure for it, so if you can guide me through it - I would highly appreciate it.

If you have any questions regarding the code - I'm ready to answer.


marcozaro:
Hi Anton,
can you please explain a bit the new functionalities you introduced?
thanks

marco

AntonSafr96:
@marcozaro  
> Hi Anton, can you please explain a bit the new functionalities you introduced? thanks
> 
> marco
Dear Marco,

It works in a following way:
We take a formula for the cross-section in the collinear factorization approach, where two main parts of it are PDFs of the incoming particles.
In MG5 these two functions are represented by two `LHAIDs` in the run_card.
So, if user needs asymmetric collisions, he should specify two different PDFs (and change the corresponding flag to `True` state). Then we are going to calculate two default cross-sections, let's say `sigma_AA ` and `sigma_BB` (where A and B are two different incoming particles, with two different `LHAIDs`) plus two additional cross-sections - `sigma_AB` and `sigma_BA`.

In my approach - I modified `parton_lum_x.f` files (you can find them in every folder after generation) in a way that, using two additional arrays, I can store information about what partons are involved in the collision, and then using the call of the LHAPDF library, I can match these partons with different PDFs (it's done in the fks_singular.f) and construct the necessary cross-sections. 

Since this "merging" process is done in two functions (`reweight_pdf` and `reweight_scale`) it gives us the possibility to run the PDF and scale variations (together with the calculation of the PDF and scale uncertainties) for our new `sigma_AB` and `sigma_BA`. 

A small remark here: There is a simplification that I have used in my approach. In the construction of a new cross-section, I take only the central value of the `PDF_A`. So, if I need to make PDF variations - it would be done only from the side of the second PDF, so, in other words, it would be the second PDF uncertainty only.

As the output code is using HwU files, which means that historams.py was also modifed in order to add additional arrays of data to the final table.
Also, using histograms.py - user has additional plots for Nucler Modification Factors for all kin.functions (or histograms) that are specifed in the analysis_card.

This is a global and small scope of the work that been done. If you have additional questions or want me to explain something in details - I can do it here

rikkert-frederix:
Hello Anton,

I'm not sure what is meant by "asymmetric collisions". If it is just a different PDF set for the left and right incoming particles, why do you need sigma_AA, sigma_BB, sigma_AB and sigma_BA all at the same time? Wouldn't only one of them (e.g., sigma_AB) be enough? That is, doesn't the accelerator define what are the two incoming particles and that fixes if the left and right incoming are an 'A' or a 'B'?

Furthermore, I would have expected that for asymmetric collisions, you might also want to use different factorisation scales for the two incoming sides. Are my expectations wrong, or is this something for a future implementation?

Best,
Rikkert

AntonSafr96:
> Hello Anton,
> 
> I'm not sure what is meant by "asymmetric collisions". If it is just a different PDF set for the left and right incoming particles, why do you need sigma_AA, sigma_BB, sigma_AB and sigma_BA all at the same time? Wouldn't only one of them (e.g., sigma_AB) be enough? That is, doesn't the accelerator define what are the two incoming particles and that fixes if the left and right incoming are an 'A' or a 'B'?
> 
> Furthermore, I would have expected that for asymmetric collisions, you might also want to use different factorisation scales for the two incoming sides. Are my expectations wrong, or is this something for a future implementation?
> 
> Best, Rikkert

Dear @rikkert-frederix 
Thank you for the reply

I) Indeed, in our approach for the "asymmetric collisions"  we just use a different PDF set for the left and right incoming particles. 
The main idea behind having four cross-sections is to be able to compute Nuclear Modification Factors (like sigma_AB/sigma_AA) in one iteration (in a single run).

II) I guess you're right. But for now - we limit ourselves to the single Factorization scale for both particles (right and left). In the future, we are going to implement additional factorization scales for both parts.

madgraph/various/banner.py
```diff
@@ -868,7 +868,7 @@ def append(self, line):
                 self.info['full_model_line'] = line
                 self.clean(remove_bef_last='import', keep_switch=True,
                         allow_for_removal=['generate', 'add process', 'add model', 'output'])
-                if cmds[1] == 'model' and len(cmds)>2:
+                if cmds[1] == 'model':
```
oliviermattelaer: why removing this condition? this is making the code to crash if this is not true?

madgraph/various/banner.py
```diff
@@ -5293,7 +5293,12 @@ def default_setup(self):
         self.add_param('lpp1', 1, fortran_name='lpp(1)')        
         self.add_param('lpp2', 1, fortran_name='lpp(2)')                        
         self.add_param('ebeam1', 6500.0, fortran_name='ebeam(1)')
-        self.add_param('ebeam2', 6500.0, fortran_name='ebeam(2)')        
+        self.add_param('ebeam2', 6500.0, fortran_name='ebeam(2)')
+        
+##############Anton
+        self.add_param('asymm_choice', False, fortran_name='asymm_choice',comment='Can be True/False. Works only with reweight_scale or/and reweight_PDF flags = True ')
+##############Anton
```
oliviermattelaer, line 5325: Should not add a condition in check_consistency that if asymm_choice is on True, lhadpdf is on?
and that reweight_pdf is On also?

For the first condition, i guess it makes sense to crash in that case,
while for the second, you can switch it on with a warning/information

Template/NLO/SubProcesses/setcuts.f
```diff
@@ -416,7 +416,11 @@ subroutine set_tau_min()
                endif
                xw(i)=0d0
             enddo
-            stot = 4d0*ebeam(1)*ebeam(2)
+            if (ebeam(1).le.0.938.or.ebeam(2).le.0.938) then 
+                stot=2d0*ebeam(1)*ebeam(2)
+            else 
+                stot=4d0*ebeam(1)*ebeam(2)
+            endif
```
oliviermattelaer: This need at least a comment to explain "why", you change the formula...
Should not we use the correct formula (without any assumptions)?

Template/NLO/SubProcesses/reweight_xsec_events.f
```diff
@@ -499,57 +519,150 @@ subroutine reweight_scale_ext
       return
       end
 
-      
+!<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<AntonS
```
oliviermattelaer: Just a style detail
avoid to use "<<<<<<<" since those are use for marking conflict by git (or other)

Template/NLO/SubProcesses/genps_fks.f
```diff
@@ -945,7 +945,11 @@ subroutine generate_momenta_conf(input_granny_m2,ndim,jac,x
       enddo
       if( firsttime .or. iconfig0.ne.iconfigsave ) then
          if (nincoming.eq.2) then
-            stot = 4d0*ebeam(1)*ebeam(2)
+            if (ebeam(1).le.0.938.or.ebeam(2).le.0.938) then 
+                stot=2d0*ebeam(1)*ebeam(2)
+            else 
+                stot=4d0*ebeam(1)*ebeam(2)
+            endif
```
oliviermattelaer: same here, please add comment on those lines

Template/NLO/Cards/run_card.dat
```diff
@@ -66,6 +66,18 @@
  %(ebeam1)s   = ebeam1  ! beam 1 energy in GeV
  %(ebeam2)s   = ebeam2  ! beam 2 energy in GeV
 #***********************************************************************
+# Choice of the collision system:
+#    True/False = asymmetric / symmetric
+#
```
oliviermattelaer, line 81: Should not you rather comment on what it does  (and how)?
Otherwise people will be confuse

oliviermattelaer (changes requested):
In order to put pressure on Rik, Marco to release 3.6.0, let me do a first review (mainly on the python part). See the small comment below.

One global comment, Is that I would have expect some change in various file to fix the previous assumptions that lab-frame = center of mass of the hadronic system.
Are you not sensitive to that or do I miss those changes?

Cheers and thanks,

Olivier

madgraph/various/banner.py
AntonSafr96: Hi Olivier,

I didn't touch this line, but maybe it's something new, I have never seen it.
But I can tell that in 3.x this line is there - so I'm going to add it.

Template/NLO/SubProcesses/setcuts.f
AntonSafr96: If you won't change that, then the global sqrt(s) of the process will be calculated using factror 4 and not 2, and case of fixed-target - it is not correct. It the same for this fine and for genps_fks.f

Template/NLO/Cards/run_card.dat
AntonSafr96: I agree. I will add more details about this “asymmetry” case.

rikkert-frederix:
Hello Anton,

I'm a bit concerned about your implementation. The reason is the following. The main idea/goal of this branch, i.e., allowing for a different PDF for the left and right incoming particles, I fully support and it would be a great feature to have. However, I'm not so sure that the current way of getting there is the best way. It seems overly complicated, with several "hacks" (loops to 'jmax’ with almost identical codes) that will be difficult to maintain/support in the future when/if new additions are made to the MadGraph code. In my opinion, the implementation should and can be kept much simpler.

Currently (in the main code), there is an array 'lhaPDFid' that list which PDF sets should be included in the reweighting. The most straight-forward way of having separate PDF sets for left- and right-incoming beams is to make this a 2D array. Then, in the parton-lum* files, before the luminosity of the 2nd beam is computed, a check can be made to see if the PDF sets for the left and right beams are identical, and, if not, a call to initPDFm is made to update the sets. (And also after the 2nd beam’s luminosity is computed, one can re-set the PDF set to the original value). I believe this would be a much simpler implementation that only requires minimal changes. In this case, no major changes are needed anywhere besides this file. All other changes would just be to pass the 2-D lhaPDFid array correctly to either the event writing and plotting routines, and there is no need to have complicated loops with "jmax" etc: all weights will be correctly dealt with from the start. What do you think about this? Could it work?

Furthermore, you’ll need several checks in the code:

1. the value of the strong coupling needs to be taken consistently with the PDF — at least for pp collisions. I’m not sure what happens for pA collisions and which PDF set should be used to provide the strong coupling. Maybe some check is needed that alphaS from either set has the same central value and runs in exactly the same way?

2. What happens if one includes PDF uncertainties? Can this be done consistently only for the left- or right-incoming particles? Does the code do this correctly?

3. Is the input in the run_card consistent? Does one put the right number of PDF sets. Probably also here one needs to allow for a 2-D array for the 'lhaid' value, and this needs to be correctly interpreted by the banner.py code. 

Also, I’m not sure that for fixed target experiments the whole setup (either at fNLO or with MC@NLO matching) of the code (PS mappings, etc.) is correct and one does not end up with non-physical situations. For example beams moving in the wrong directions after boosting, etc. I guess this is related to the fact that the collinear factorisation used is (formally) only compatible with massless initial states. I assume that you have all this worked out and it can work?

Anyway, if you want, we could have a chat over zoom to discuss a bit more. Maybe I’ve missed something important and my suggestion cannot work. In that case, write me an e-mail.

Best,
Rikkert

stahlleiton:
What is the status of the use of asymmetric collisions in MadGraph5?

PR #85, opened 2024-03-11
fixing grouping issue for p p > a a j [QCD]

oliviermattelaer:
This include two change
1) the change of the operator of comparison for the loop
(from the equality of virt_me_matrix to the comparison of the METag)

2) change in the model restriction.
The reason is that loop-matrix element generation was keeping lorentz structure that was not associated to a coupling (or more exactly coupling dropped by the model due to the restriction). Therefore by also removing the lorentz structure from the model, this change the loop representation and therefore the METag/diagram representation.

Note that such change has no impact on the generated code and/or value of the matrix-element


madgraph/fks/fks_helas_objects.py
```diff
@@ -834,7 +834,14 @@ def __eq__(self, other):
...
         # now the virtuals
-        if self.virt_matrix_element != other.virt_matrix_element:
+        if self.virt_matrix_element and other.virt_matrix_element: 
+            virttag = helas_objects.IdentifyMETag.\
+                        create_tag(self.virt_matrix_element.get('base_amplitude'))
+            othertag = helas_objects.IdentifyMETag.\
+                        create_tag(other.virt_matrix_element.get('base_amplitude'))
+            if virttag != othertag: 
+                return False
+        elif self.virt_matrix_element !=  other.virt_matrix_element:
```
marcozaro, line 844: in practice, the code arrives here if either of the two virtual is empty, right?

madgraph/interface/common_run_interface.py
```diff
@@ -7587,7 +7587,7 @@ def open_file(self, answer):
                 answer = 'plot'
             else:
                 answer = self.cards[int(answer)-self.integer_bias]
-
+        path = ''
```
marcozaro, line 7586: this is unrelated to the issue at hand, right?

models/import_ufo.py
```diff
@@ -2514,6 +2514,26 @@ def remove_interactions(self, zero_couplings):
                                  ' with loop particles (%s)'%loop_parts+\
                                  ' perturbing order %s'%order)  
 
+        # looping over all vertex and remove all link to lorentz structure that are not used anymore
```
marcozaro, line 2517: not sure to understand this chunk of code...

madgraph/fks/fks_helas_objects.py
oliviermattelaer: yes correct

madgraph/interface/common_run_interface.py
oliviermattelaer: yes it is uncorrelated, this fixed another issue in another branch and made a mistake to include this in this PR (but would prefer to keep it anyway)

models/import_ufo.py
oliviermattelaer: So originally (before restriction) the vertices  are like this 
color = [c1,c2, ...] 
lorentz = [l1,l2,....]
coupling = {(i1,j1): C1, (i2,j2): C2, ....}

When running the restriction, some coupling are dropped from the model.
But before this patch, neither the lorentz list nor the color list were modified (so just the coupling list was made shorter).

So the issue (only for loop) is that those "not used" lorentz structure was kept in the comparison operator and therefore they were kept separated. 

So this code is cleaning the lorentz structure and changing the coupling part since the index are modified when I'm doing that.

marcozaro: ok. is there any unittest checking this, just to be sure not to break it in future changes?

oliviermattelaer: good point. I was lazy on that.

They were already  a unittest related to that function, but obviously was not focussing on that point.
So I have added two unittest to test one "trivial" case and a second less trivial. 
By limiting myself to SM case (for speed), I do not  deeply test most complex case that can be generated but I think that this is enough.

I have also run those unittest on 2.9 version, and they indeed fail (as expected).

marcozaro:
Then OK for me for merging

oliviermattelaer:
Thanks Marco.

PR #87, opened 2024-03-25
Update input events when calculating maxwgt in get_maxwgt_for_onshell

sargyro:
In `get_maxwgt_for_onshell`, `orig_lhe.seek(0)` was called from inside the loop over the input events and as a result, the input events were never updated (instead of probing the first 75 events, only the 1st event was probed). 

I move the `orig_lhe.seek(0)` outside of the loop, so that the input events are updated.

@oliviermattelaer maybe you also propagate this to the main branch, since it's also affected?


oliviermattelaer:
Thanks a lot,

Yes will double check that in the official version.

Olivier

PR #88, opened 2024-04-04
add hepmc and fastjet libs to execute rivet

sihyunjeon:
nothing much to add,
somehow for some reason hepmc and fastjet .so were not taken when executing rivet in madgraph
this PR just makes the subscript to run rivet to include those .so files

work as intended (yoda files are being produced without complaints)


oliviermattelaer:
Perfect let's approve this :-)
Thanks so much

PR #91, opened 2024-04-11
Support for the FD gauge

oliviermattelaer:
- adding a new gauge "set gauge FD"
- For this the model need to support the Feynman gauge and if it does the FD is constructed on the flight (by merging interaction/...)


oliviermattelaer:
Hi Marco,

Any objection on this? I would like to move forward on 3.6.0.

Cheers,

Olivier

aloha/aloha_writers.py
```diff
@@ -536,7 +543,11 @@ def get_declaration_txt(self):
         if KERNEL.has_pi:
             out.write(' parameter (PI=%s)\n' % self.change_number_format(cmath.pi))
         
-        
+        if aloha.unitary_gauge == 3: # Feynman
```
marcozaro: Feynman->FD gauge?

bin/create_release.py
```diff
@@ -172,7 +172,14 @@
     auto_update = False
 
 # checking that the rev_nb is in a reasonable range compare to the old one.
```
marcozaro: there are some conflicts

madgraph/core/color_algebra.py
```diff
@@ -957,6 +957,8 @@ def to_canonical(self, immutable=None):
     def __eq__(self, col_str):
         """Check if two color strings are equivalent by checking if their
         canonical representations and the coefficients are equal."""
+        #misc.sprint(self.coeff, self.Nc_power, self.is_imaginary, self.to_canonical())
```
marcozaro, line 960: remove these two lines?

madgraph/interface/amcatnlo_run_interface.py
```diff
@@ -5272,6 +5272,74 @@ def copy_lep_densities(self, name, sourcedir):
             files.cp(os.path.join(lep_d_path, filename), pdf_path)
 
 
+    def link_and_copy_epdf(self, pdlabel, lhaid, libdir):
```
marcozaro: why this change here? unrelated to the FD gauge, right?

marcozaro (changes requested):
generally ok! nice work, see a bunch of minor comments

aloha/aloha_writers.py
oliviermattelaer: thanks

bin/create_release.py
oliviermattelaer: -> fix

madgraph/core/color_algebra.py
oliviermattelaer: That one I would prefer to keep since they are helpfull for debugging

madgraph/interface/amcatnlo_run_interface.py
oliviermattelaer: Looks like this is a git error (or when solving a conflict), since now this function is define twice within this file (line 5207 and 5275)... Thanks for spotting this.

oliviermattelaer (commented):
Thanks Marco,

I have implement all the change that you spot (thanks so much for detecting that I mess a up on one of merging of this branch). I have just not act on your request to remove a comment since I find it usefull for debugging if/when needed.

Thanks,

Olivier

marcozaro (approved):

marcozaro:
great!

Marco

PR #94, opened 2024-04-18
Patch to running plugins from the PYTHONPATH

zlmarshall:
Two minor patches to the plugin setup:

- First, if a plugin isn't installed, the code crashes in a rather non-obvious way. This changes it to simply sys.exit after the plugin has not been found and an error message has been printed.

- Second, in case a plugin is brought in via the PYTHONPATH (MG5aMC_PLUGIN), it needs to then be loaded from there with the sys module. The current setup leads to it not being found in the module dictionary.

I hope these two are clear enough, but as this is my first pull request please let me know if there's anything else I should do here, any other documentation that would be useful, tests to run, etc.

### DIFF
--- a/bin/mg5_aMC
+++ b/bin/mg5_aMC
@@ -161,12 +161,13 @@ if options.plugin:
     if not os.path.exists(os.path.join(root_path, 'PLUGIN', options.plugin)):
         try:
             __import__('MG5aMC_PLUGIN.%s' % options.plugin)
+            plugin = sys.modules['MG5aMC_PLUGIN.%s' % options.plugin]
         except:
             print( "ERROR: %s is not present in the PLUGIN directory. Please install it first" % options.plugin)
-
+            sys.exit()
     else:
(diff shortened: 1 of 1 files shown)


oliviermattelaer:
For me this is good to go, thanks for this

The only question is if this is not the kind of patch that would make sense for the LTS version?
(If you agree then I will cherry-pick your commit and push it in the 2.9.20 branch (--I hope you will keep paternity--)

Cheers,

Olivier

zlmarshall:
Thanks @oliviermattelaer ! No objection to this being cherry-picked into 2.9.20! (And no issue if my name disappears from it, of course, I'm not that kind of "developer" :) ).

Best,
Zach

oliviermattelaer:
Finally, I did not pushed this to 2.9.20 since the feature is actually not supported at all with that version. 
So this would have been equivalent to add a feature which is forbidden for the LTS.

So this simplify things and this is merged into 3.6.0.

Thanks,

Olivier

PR #100, opened 2024-06-03
Fix moved cern amcatnlo webpage

APN-Pucky:
I have not looked for more cern.ch pages that might also need a changed URL.

### DIFF
--- a/Template/NLO/Cards/run_card.dat
+++ b/Template/NLO/Cards/run_card.dat
@@ -134,7 +134,8 @@
 #   0: No merging                                                      *
 #   3: FxFx Merging - WARNING! Applies merging only at the hard-event  *
 #      level. After showering an MLM-type merging should be applied as *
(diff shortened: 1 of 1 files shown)


oliviermattelaer:
Thanks for the notification, I will ask the website responsible if this is a temporary change or not and then see what is the correct action to perform.

Cheers,

Olivier

APN-Pucky:
Thanks, any update? As of now both sides are unavailable...

oliviermattelaer:
Information last month was that yes we had to update the link, but looks like the cern transition out of afs is creating additional issue. I did send an email to the website responsible, and they will likely opening a ticket to cern IT to see what the status is.

Thanks for the notification,

Olivier

oliviermattelaer:
The website https://amcatnlo.web.cern.ch/amcatnlo/
and https://amcatnlo.web.cern.ch
are now both responding (and both point to the same content)

Thanks,

Olivier

PR #101, opened 2024-06-06
No integration channels -> suggest masses too high or beam energy too low

APN-Pucky:
I think if there are no integration channels the user should also be suggested/warned that maybe the beam energy is too low or masses too large. Because after the "No integration channels..." warning all I got was a ArrayIndex out of bounds error on the empty jobs array.

In the end I found out by manually running `gensym` to see why there is no `channels.txt` and it told me about the energy.
image

I wasted a bit too much time wondering if there is an error with a specific model (http://feynrules.irmp.ucl.ac.be/wiki/MSSMatNLO) or specific MadGraph versions because of this.


oliviermattelaer:
Thanks a lot,

Olivier

PR #102, opened 2024-06-07
Patch to indentation in plugin handling

zlmarshall:
It looks like commit 7c52d7d491baf12ee0aa49356cec3dc8e68407eb introduced a bug in the plugin handling. In the PR over here:

https://github.com/mg5amcnlo/mg5amcnlo/pull/94/files

The plugin line is indented by four spaces. That's very important - we only want to redefine plugin using PLUGIN.blah in the case that it's imported that way. Otherwise it should be defined as MG5aMC_PLUGIN.blah, as it is a few lines above.

This PR adds the spaces back in.

### DIFF
--- a/bin/mg5_aMC
+++ b/bin/mg5_aMC
@@ -161,9 +161,10 @@ if options.plugin:
     if not os.path.exists(os.path.join(root_path, 'PLUGIN', options.plugin)):
         try:
             __import__('MG5aMC_PLUGIN.%s' % options.plugin)
+            plugin = sys.modules['MG5aMC_PLUGIN.%s' % options.plugin]
         except:
             print( "ERROR: %s is not present in the PLUGIN directory. Please install it first" % options.plugin)
-
+            sys.exit()
     else:
         __import__('PLUGIN.%s' % options.plugin)
         plugin = sys.modules['PLUGIN.%s' % options.plugin]
--- a/bin/mg5_aMC
+++ b/bin/mg5_aMC
@@ -166,7 +166,7 @@ if options.plugin:
 
     else:
         __import__('PLUGIN.%s' % options.plugin)
-    plugin = sys.modules['PLUGIN.%s' % options.plugin]
+        plugin = sys.modules['PLUGIN.%s' % options.plugin]
     if not plugin.new_interface:
         logging.warning("Plugin: %s do not define dedicated interface and should be used without the --mode options" % options.plugin)
         sys.exit()


oliviermattelaer:
But you would need also a line
plugin = sys.modules[''MG5aMC_PLUGIN.%s' % options.plugin]

no?

zlmarshall:
Absolutely! I've added that. It was also a part of https://github.com/mg5amcnlo/mg5amcnlo/pull/94/files , so I'm not sure what went wrong there — the whole PR was accepted but none of the code changes actually appear in the branch? Maybe it was reverted?

Thanks,
Zach

oliviermattelaer:
One reason is that such commit was included within 3.6.0 branch which is not yet merged within 3.x branch
So I did change the target to 3.6.0 for this PR.

But I'm still confused by how git is handling this
On 3.6.0 the content of those lines sounds reasonable:
https://github.com/mg5amcnlo/mg5amcnlo/blob/2b25c97f9863ee2c8a934af1855427677fed543f/bin/mg5_aMC#L160

But you still have a difference with those in this PR.... (and no conflict reported, which I do not understand)

Do you agree with me that 3.6.0 branch is correctly handling this?
If yes, I will just close this PR (or I might try to merge it anyway. Just to see if git behaves correctly or not)

zlmarshall:
Ahhhh, I see. I hadn't realized you had this additional branch structure (our standard is to not use an all-numbers name until the release exists; so e.g. 3.6.0 would only exist as a branch when it was made as a release from 3.6.x or 3.x). You might think about automatic sweeps to keep some of the development lines in sync. In any case: yes, 3.6.0 seems to do the right thing. This PR originally targeted 3.x (just like my last one did), and I suspect the diff wasn't updated and won't be unless I rebase my changes. I think it's ok to close this, then, and I'll wait patiently for 3.6.0. Do you have a target date for that release? (Or maybe not 'day' but 'number of weeks / months'?)

Thanks,
Zach

oliviermattelaer:
Perfect thanks, 

(to reassure me, I did the merge on my local machine and the merge commit is identical to 3.6.0).
So I will close this PR.

For the target date of 3.6.0, they are still two "big" PR that I would like to merge within 3.6.0 before the release (maybe three). One of them (sudakov) is fully ready but the second (allowing cuda/simd and openmp) still need some clean up to be applied to avoid side effects on the rest of the code. But this is a matter of a couple of weeks now.

Cheers,

Olivier

PR #105, opened 2024-06-12
Gpucpp fix scan vector

oliviermattelaer:
This was ready ... at least before the merge of the wrap branch
So I put this as WIP for the moment to check the status


oliviermattelaer:
OK  this is good to be merged as the changed here are orthogonal to the change to the upstream repo for the wrap.
The associated PR in the plugin will be more complex to handle.

oliviermattelaer:
Note this also include: https://github.com/mg5amcnlo/mg5amcnlo/tree/gpucpp_826
(which ACTUALLY needs to be secure if the plugin does not provide the new API)
-> will fix this here

PR #106, opened 2024-06-12
allow for second exporter to have access to all variable used in the …

oliviermattelaer:
Likely already merged but this will allow to check it


oliviermattelaer:
Yes this is include in the other PR https://github.com/mg5amcnlo/mg5amcnlo/pull/105
But this current PR is not perfect (needs protection if the plugin does not have define the new API ...
-> will fix that in the other PR, so just close this one

PR #107, opened 2024-06-18
Parallelize untar operation in case of gridpack

Qubitol:
This PR improves the untarring of the zip files created when using the `gridpack` option, by parallelizing the operation using `xargs`.

On a separate, but related, issue, the unzipping is performed by following lines of `bin/internal/restore_data`:
```bash
gunzip \$j
cp $1_ftn26 ftn26
gzip $1_ftn26
```
However, re-zipping the file can be avoided by using the `--keep` option of `gunzip`, maybe we can save even more time.

### DIFF
--- a/Template/LO/bin/internal/restore_data
+++ b/Template/LO/bin/internal/restore_data
@@ -45,29 +45,25 @@ for i in `cat subproc.mg` ; do
     else
 	cp  results.dat $1_results.dat  >& /dev/null
     fi
-    for k in G* ; do
-	if [[ ! -d $k ]]; then
-	    continue
-	fi
-	cd $k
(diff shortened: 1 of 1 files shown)


oliviermattelaer:
Hi Daniele,

Yes you can use the --keep this is a very good idea!.

Then this is good to merge.

Thanks a lot,

Olivier

Qubitol:
Hi Olivier,

I added the `--keep` and changed the copy to an `mv` in the line next to it.
If you agree, we can merge, but apparently I don't have write access to the repo.

Thanks.

Daniele

oliviermattelaer:
Thanks, I have done it myself then :-)

PR #115, opened 2024-06-28
Fix icolamp in export_cpp.py (against gpucpp_june24?)

### LINKED ISSUE #114: coloramps differences in fortran and cpp
Hi @oliviermattelaer this is a followup to https://github.com/madgraph5/madgraph4gpu/pull/873 and related issues.

I hope I will manage to idetify the differences in f77 and cpp.
There is one thing anyway that puzzles me, feel free to close as irrelevant

The f77 version has
```
                if (diag_tuple[4]-diag_tuple[5]) == max_Nc:
                    diag_num = diag_tuple[0] + 1
                    # Add this JAMP number to this diag_num
                    diag_jamp[diag_num] = diag_jamp.setdefault(diag_num, []) + \
                                          [ijamp+1]
                else:
                    self.proc_characteristic['single_color'] = False
```
In the cpp version, those two lines are commentd out. Why?

And then there are other subtle differences I am loking at.
Thanks
Andrea

valassi:
Hi @oliviermattelaer finally this is a MR for fixing the icolamp issue (discussed in #114 but also much more extensively in madgraph5/madgraph4gpu#873)

Can you please review?

Thanks
Andrea

PS Note, in cudacpp we had TWO separate issues in the coloramps.h file, contributing to madgraph5/madgraph4gpu#856: one is the one you worked on in madgraph5/madgraph4gpu#852 about the mapping of iconfig to channelid, for which you added an array in coloramps.h; but then a second independent issue is that the icolamp array in coloramps.h was also wrong. This second issue is what is fixed here.


oliviermattelaer (changes requested):
The goodjet should not be include at this stage. 
This is not yet understood.

The rest is ok (but one can clean the code a bit by removing debug statement but that is not really critical)

oliviermattelaer:
So goodjet change can actually be include since it has been tested now.
But since #116 is currently waiting proper checking on the change of one line, the PR should have the same status.
We should simply close this PR and just merge (or better do a PR) gpucpp into gpucpp_june24 such that gpucpp_june24 is correctly up-to-date and can continue to move forward on the goal to be merged within 3.6.0.

valassi:
Note; these issues are fixed in https://github.com/madgraph5/madgraph4gpu/pull/880 where Olivier provides a much better fix.

Thanks Olivier!

PR #117, opened 2024-07-02
Gpucpp

oliviermattelaer:
This is to double check what was pushed in gpucpp and to sync it with gpucpp_june24 (and check if they are issue introduced within the gpucpp branch)

### DIFF
--- a/Template/LO/SubProcesses/reweight.f
+++ b/Template/LO/SubProcesses/reweight.f
@@ -620,6 +620,7 @@ C   local variables
 
 c     Variables for keeping track of jets
       logical goodjet(n_max_cl)
+c111      logical set_goodjet(n_max_cl) ! variable for debugging unset call 
       integer fsnum(2),ida(2),imo,jcode
       logical chclusold,fail,increasecode
       save chclusold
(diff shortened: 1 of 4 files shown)


Template/LO/SubProcesses/reweight.f
```diff
@@ -633,6 +633,17 @@ logical function setclscales(p, keepq2bck, ivec)
...
+
+c Workaround for valgrind 'Conditional jump or move depends on uninitialised value(s)'
+c See https://github.com/mg5amcnlo/mg5amcnlo/issues/111
+c FIXME: this is just a workaround to avoid uninitialised values and undefined behaviour...
+c FIXME: a real bug is probably hidden in the code (goodjet must be correctly defined!)...
+c FIXME: adding this workaround may change the behaviour of existing code...
+      do i=1,n_max_cl
+        goodjet(i)=.false. !!FIXME!! there is no reason to choose false instead of true here...
+      end do
+
```
oliviermattelaer:
```suggestion
     setclscales=.true.
c    ensure that goodjet start on False to avoid potential undefined behavior since the algorithm           
c    does call unset variable. The associated behavior was detected stable due to the presence of the 
c   requirement of other condition in the same if statement. In case of weird behavior, it is adivsed to add 
c   a second array to check which element are set from those which are not      
         goodjet(:) = .false.
```

Template/LO/SubProcesses/reweight.f
```diff
@@ -1227,7 +1238,7 @@ logical function setclscales(p, keepq2bck, ivec)
          do i=1,2
             do j=1,2
 c              First adjust goodjet based on iqjets
-               if(goodjet(ida(i)).and.ipart(j,ida(i)).gt.2)then
+               if(goodjet(ida(i)).and.ipart(j,ida(i)).gt.2)then ! FIXME: goodjet is uninitialised #111
```
oliviermattelaer:
```suggestion
               if(goodjet(ida(i)).and.ipart(j,ida(i)).gt.2)then ! DEBUG: goodjet() can access here not set variable
```

oliviermattelaer (commented):
Ok this is good to be merged.
Just change the part associated to goodjet since we should remove those FIXME

oliviermattelaer:
Ok the diff was useful but allowing github to merge the conflict induced some unwanted side effect.
Sorry for that. I will do such merge by hand bypassing the PR to avoid the issue

PR #124, opened 2024-08-02
Goodhel

oliviermattelaer:
Change madevent handling of the helicity for non helicity recycling case
- the list of helicity is now the same independently of imirror (no reason to have them different)
- the list of helicity is now different for each of the matrix-element (here this makes a lot of sense to distinguish u u versus u u~) 
- this_ntry is not needed anymore (since ntry depends of the grouping)


oliviermattelaer:
dedicated check:
 - loop induced p g > h j has been checked and event are 100% identical with and without such change.

oliviermattelaer:
Just to add that this also add limhel as a parameter of the run_card.dat

PR #125, opened 2024-08-02
do not remove GpuAbstraction.h in madevent_interface.py

valassi:
Hi @oliviermattelaer as discussed in https://github.com/madgraph5/madgraph4gpu/issues/947

- I added one line in madevent_interface.py to avoid deleting a G* path if it is not a directory (as you suggested - this will avoid deleting GpuAbstraction.h in particular)
- I moved this here upstream so that it is not in a patch

Can you please review? Hopefully not controversial

Thanks
Andrea


valassi:
PS I also just added a comment in the code mentioning issue 947

oliviermattelaer:
Thanks,

Olivier

valassi:
Very good thanks to you :-)

PR #121, opened 2024-07-19
master_june24 completion: nb_warp_used (and a SUBset of gpucpp_june24) into gpucpp

valassi:
Hi @oliviermattelaer this is a WIP MR into gpucpp of my nb_warp_used and of everything else I am using against master_june24 in my (https://github.com/madgraph5/madgraph4gpu/pull/882).

The nb_warp_used fixes are described here https://github.com/madgraph5/madgraph4gpu/issues/885#issuecomment-2233827188

NB: as discussed in https://github.com/madgraph5/madgraph4gpu/issues/886, gpucpp_june24, with or without my changes, does NOT work for me with master_june24 (my modified version, but I am pretty sure that the problem was pre-existing). This is why here I am NOT merging gpucpp_june24, I am only merging a SUBSET of that (what was originally used in master_june24, plus my changes, see https://github.com/valassi/mg5amcnlo/commits/valassi_gpucpp_june24/

I keep this in WIP for the moment. I would only merge this when we agree to merge my https://github.com/madgraph5/madgraph4gpu/pull/882 into master (as master is meant to use the head of gpucpp... I do not want to 'pollute' gpucpp before!).

I also filed PR #120 against gpucpp_june24, but as mentioned in https://github.com/madgraph5/madgraph4gpu/issues/886, **I am unable to use gpucpp_june24 and this is NOT what was used in master_june24** anyway.

Thanks Andrea

PR #126, opened 2024-08-06
Gpucpp goodhel

oliviermattelaer:
This is the equivalent PR for merging the goodhel support within gpucpp branch
This includes  the update of the madgraph version to 3.6.0 (madgraph itself is here still untested, the full support is for another PR)
which in particular includes
 - limhel being now a user parameter
 - goodhel being now refactored to limit call to reset function. It additionally does not use the imirror distinction but use a NPROCESS distinction now (but currently NPROCESS=1 at the fortran level at least)

This should allow to fix the failing CI about missmatch in cross-section due to different number of call associated to the reset function.


oliviermattelaer:
Hi @andrea,

So this PR is getting obsolete (or will be very soon) since this branch is now merged in the much more ambitious gpucpp_for360 branch.
The only potential "downside" is that gpucpp_for360 was staring from gpucpp_june24 which associated PR could also be consider as obsolete to focus only on this one.

So in short  gpucpp_for360 will contain 
 - current gpucpp, 
 - gpucpp_june24 (in particular the warp support at fortran level)
 - gpucpp_goodhel (this one)
 - latest 3.6.0 (since this PR contains 3.6.0
 - and obviously new content for gpucpp_for360 (took me 45 days of work --not full time--) which is mainly fixing many bug in MG5aMC and therefore have to CI on beautifull green.

So I would propose that we close all those PR to focus only on the one gpucpp_for360.

valassi:
> Hi @Andrea,
> 
> So this PR is getting obsolete (or will be very soon) since this branch is now merged in the much more ambitious gpucpp_for360 branch. The only potential "downside" is that gpucpp_for360 was staring from gpucpp_june24 which associated PR could also be consider as obsolete to focus only on this one.
> 
> So in short gpucpp_for360 will contain
> 
>     * current gpucpp,
> 
>     * gpucpp_june24 (in particular the warp support at fortran level)
> 
>     * gpucpp_goodhel (this one)
> 
>     * latest 3.6.0 (since this PR contains 3.6.0
> 
>     * and obviously new content for gpucpp_for360 (took me 45 days of work --not full time--) which is mainly fixing many bug in MG5aMC and therefore have to CI on beautifull green.
> 
> 
> So I would propose that we close all those PR to focus only on the one gpucpp_for360.

Hi @oliviermattelaer thanks and sorry for the delay.

I would prefer to go in steps, especially the june24 and goodhel chunks are big.

I suggest that we do either the goodhel or june24 tonight, then I fix the other one.

And when we have merged both chunks, probably what remains is what you call "new content for 360", which I guess is a lot...

(SOME of the 360 stuff, in any case, is in what I call the goodhel chunk, ie this very #126)

NB: for me this #126 is good to go, but first we need to merge #137 into it. Can I go aheads with that?

oliviermattelaer:
Hi, actually this PR is only big because it contains the update from 3.5.4 to 3.6.0 (and 3.6.0 is a BIG BIG release).
So in itself this PR is super light. But indeed merging this goodhel with june24 is a lot of work... which is (mostly) actually done in gpucpp_for360.

So (as said by email before reading this thread) I would (at least try) to not duplicate such (big) work of merging by moving directly to gpucpp_for360 which in itself does not contains new features (but many many bug fix on the mg5amc side).
Can we give it a try?  I think that moving to gpucpp_for360 will allow to avoid to duplicate complicated merge fixing and will save us time.

valassi:
I have merged my valassi_goodhel here in PR #137.

Note that this includes the pre-june24 status of gpucpp. So, now that we updated gpucpp with june24 in #121, this is not mergeable into gpucpp as-is.

I am preparing another (very WIP) PR for allowing the merge with june24.

valassi:
Hi @oliviermattelaer I guess that we can close this one and instead merge #140 ?

PR #131, opened 2024-08-23
add compile flags for gcc14 compilation

joequant:
This disables pointer mismatch warnings which are now errors in gcc14.


oliviermattelaer:
Thanks I have cherry-pick your patch to be included in our LTS version (https://github.com/mg5amcnlo/mg5amcnlo/commit/b8a3527c25e84e9587746cf7c53f0cc93f4f95ee), then it will propagated back to all our version.

Thanks,

Olivier

PR #132, opened 2024-08-30
Change the handling on how Source/makefile is written for second exporter handling

oliviermattelaer:
OK 

This change the way the source makefile is handle which is needed in order to correctly write such makefile in presence of second exporter
This is in particular important to fix the CI related to the recompilation when vector_size is changed

This also does some side change (like updating auto-completion and cleaning) which are not controversial.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ ptest*
 *.pkl
 p_ME_*
 *.log
+.DS_Store
 input/default_run_card_lo.dat
 input/default_run_card_nlo.dat
 input/mg5_configuration.txt
--- a/madgraph/interface/madgraph_interface.py
+++ b/madgraph/interface/madgraph_interface.py
@@ -1534,7 +1534,7 @@ This will take effect only in a NEW terminal
 
         if args[0] in ['group_subprocesses']:
             if args[1].lower() not in ['false', 'true', 'auto', 'gpu']:
-                raise self.InvalidCmd('%s needs argument False, True or Auto, got %s' % \
+                raise self.InvalidCmd('%s needs argument False, True, gpu or Auto, got %s' % \
                                       (args[0], args[1]))
         if args[0] in ['ignore_six_quark_processes']:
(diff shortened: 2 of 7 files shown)


oliviermattelaer:
This is a "simple" push which need to be sync to the madgraph4gpu testsuite_only_fixed change/PR (not yet open)
which is designed to fix the ci which is currently crashing since the code compilation is not correctly handle.

valassi:
Hi @oliviermattelaer thanks for including here my fixes from #135.

I realise I made a mistake, my apologies...

I added 'additional_clean' and I had added this in ProcessExporterFortran but not ProcessExporterFortranSA. I have now added one more commit here to fix that, https://github.com/mg5amcnlo/mg5amcnlo/pull/132/commits/fa706c6960932efdfbd9355c7c61790cf33d95d8 (I am adding directly here in this branch otherwise it is a mess, hope you do not mind).

valassi (approved):
Hi @oliviermattelaer here I approve the part that I can test ie cudacpp.

Otherwise, there are very many CI failures, but they seem to be in parts that do not depend on cudacpp. Typically they are about missing dependencies on vector.inc? (One example is #136). More precisely, I do not know if they are new CI failures caused by this PR, or if they were already there.

(One CI failure was due to my own bug, and I have now fixed this, see https://github.com/mg5amcnlo/mg5amcnlo/pull/132#issuecomment-2323195636).

For what concerns the work on cudacpp, I would approve and merge this, so that we can go on with the rest. But you know better if the CI failures are blockers or not.

valassi:
Hi @oliviermattelaer as discussed offline, I will merge this.

I understand from you that the CI failures are 'normal' and that they will be fixed in the gpucpp_for360 branch
(eg https://github.com/mg5amcnlo/mg5amcnlo/issues/136#issuecomment-2323231569)

I am merging without squashing. We could then delete gpucpp_fixci (or at least not use it anymore)

Thanks Andrea

valassi (approved):
Approved. Merging as discussed

PR #133, opened 2024-08-30
Gpucpp june24 fixci

oliviermattelaer:
let combine the june24 branch and the fixci one into a single one,
such that it can be used as a based one of the madgraph4gpu PR progess


oliviermattelaer:
Let's close this, since github does not produce the correct diff (not sure how to force him to recompute the diff, but it is clearly wrong)

PR #134, opened 2024-08-30
Gpucpp june24 fixci

oliviermattelaer:
second try to avoid gitub wrong diff


oliviermattelaer:
GRRR, I do not understand how those diff can be so wrong

PR #135, opened 2024-08-31
Fixes for Source makefile (on top of Olivier's CI fixes)

valassi:
Hi @oliviermattelaer this contains my fixes for Source makefile (on top of your CI fixes)
See https://github.com/madgraph5/madgraph4gpu/pull/980

Still WIP as I have to leave. But at least it gives you the idea.
Thanks
Andrea

### DIFF
--- a/madgraph/iolibs/export_v4.py
+++ b/madgraph/iolibs/export_v4.py
@@ -741,6 +741,7 @@ param_card.inc: ../Cards/param_card.dat\n\t../bin/madevent treatcards param\n'''
                        'model':model_line,
                        'additional_dsample': '',
                        'additional_dependencies':'',
+                       'additional_clean':'',
                        'running': ''} 
 
         if self.opt['running']:
--- a/madgraph/iolibs/template_files/madevent_makefile_source
+++ b/madgraph/iolibs/template_files/madevent_makefile_source
@@ -115,7 +115,7 @@ $(LIBDIR)libiregi.a: $(IREGIDIR)
 	cd $(IREGIDIR); make
 	ln -sf ../Source/$(IREGIDIR)libiregi.a $(LIBDIR)libiregi.a
 
-cleanSource:
+cleanSource: # Clean builds: fortran in this Source
 	$(RM) *.o $(LIBRARIES) $(BINARIES)
 	cd PDF; make clean; cd ..
 	cd PDF/gammaUPC; make clean; cd ../../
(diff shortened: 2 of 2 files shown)


valassi:
Hi @oliviermattelaer this is now complete. It comes together with https://github.com/madgraph5/madgraph4gpu/pull/984

Can you review both please?

Thanks
Andrea

oliviermattelaer:
Hi,

Just to check, is it possible for you to activate the CI? Looks like your commit are not running the CI (which is annoying).
Not very relevant here since we need the gpucpp_for_360 to have the CI all green anyway. (which will have been useful here).

So I do not have any conceptual issue, but since the CI are down I have to check two points
 - [x] that normal madevent is still running
 - [ ] that madweight is generated correctly
both are related to the fact that you add an additional entry in a template and I need to be sure that all instance using that template are modified correctly (I'm confident for the first, less on the second)

Will update this, when I have done the check

oliviermattelaer:
Ok madweight does go as far as such template in this version.
So not possible to state that, so let's merge this and I will see in the gpucpp_for360 branch if this is is an issue and fix it there in case.

oliviermattelaer (approved):

oliviermattelaer:
So I merged those commit within the mg5amcnlo:gpucpp_fixci branch such that I can approve and close this PR (and have the CI triggered).

valassi:
Hi @oliviermattelaer to make it clear: in this PR the only change I ask you to review is https://github.com/mg5amcnlo/mg5amcnlo/pull/135/commits/e1f63e54b2455a02b1efdfb1aa5fbae8865ea4d7

In practice, I split the 'clean' target by adding a 'cleanSource' target
- If you do `make clean` you obtain exactly the same result as before
- But in addition you can also just type `make cleanSource` (which is useful to avoid a loop through P* directories, as in cudacpp I need in any case to loop over P* directories later)

Apart from this, I only added some comments to clean and cleanSource targets.

But otherwise this is identical to what you used in https://github.com/madgraph5/madgraph4gpu/pull/980

valassi:
PS Ah sorry somehow I did not see that you had merged this.

Ok very good, merged in your gpucpp_fixci, that's good, thanks :-)

I will check the rest now

valassi:
> But otherwise this is identical to what you used in [madgraph5/madgraph4gpu#980](https://github.com/madgraph5/madgraph4gpu/pull/980)

JUst for the record: no sorry I had also moved the targets at the end through 'additional_clean'. And I defined this in one place in export_v4.py, but I should have done it in two places. Now fixed in https://github.com/mg5amcnlo/mg5amcnlo/pull/132/commits/fa706c6960932efdfbd9355c7c61790cf33d95d8 (see https://github.com/mg5amcnlo/mg5amcnlo/pull/132#issuecomment-2323195636)

PR #137, opened 2024-09-02
Update and complete gpucpp_goodhel (merge with pre-june24 gpucpp) including at least some 3.6.0 update

valassi:
Hi @oliviermattelaer of course I was forgetting. For the goodhel we ALSO need to do this on the mg5amcnlo side.

This is essentially the same as your #126, but on top it has the sync to gpucpp

I would suggest merging this into gpucpp instead of #126 

Can we go ahead?

NB both this and june24 are BIG PRs, so after merging one I will need to fix the orther, whichever the order. Let me know what you prefer...

Thanks
Andrea


valassi:
I take it back: I changed the base of this to gpucpp_goodhel, so you can see the diffs.

I suggest
- merging this #137 into gpucpp_goodhel (ASAP)
- merging then #126 gpucpp_goodhel into gpucpp (this can be done after june24 if needed)
- and then merging https://github.com/madgraph5/madgraph4gpu/pull/955 on the cudacpp side

oliviermattelaer:
I guess this PR is basically the same as #126 if we were cliking on "rebase".
In any case, no issue for me on this one.

For the two others, no objection on my side either. And no preference in the ordering (the only one would be "political" so irrelevant). So I let you choose.

Thanks,

Olivier

valassi:
Hi @oliviermattelaer thanks.

I am getting lost myself with too many branches, so I will merge this one.

Note: gpucpp has changed in the meantime via #121 (including the june24 stuff), so in any case gpucpp_goodhel will NOT be directly mergeable to gpucpp. We will need yet another one (that I am preparing).

Also note, this same valassi_goodhel has been merged to your gpucpp_for360 in #138.

Renaming this to mention pre-june24, and merging this

PR #138, opened 2024-09-03
WIP merge goodhel functionalities into gpucpp_360

valassi:
Hi @oliviermattelaer as discussed offline

I have created a valassi_goodhel_for360 branch, based on the valassi_goodhel branch.

The idea is to merge the valassi_goodhel branch into gpucpp (see #137). However I will first merge june24 into gpucpp and I will need to update valassi_goodhel accordingly.

Instead, this valassi_goodhel_for360 was frozen before the addition of june24.

In the meantime, in any case, I also gave it a try at actually fixing the conflicts with gpucpp_for360. There were three conflicts, and I fixed them as described here https://github.com/mg5amcnlo/mg5amcnlo/commit/ef1150819eb6217df3d06856a4d592b1fcecf177
```
Fix conflicts:
- .gitignore (keep tests/status in .gitignore, as in the 360 version)
- madgraph/iolibs/export_v4.py (keep the gpucpp_for360 version, which is more recent)
- models/template_files/fortran/makefile_standalone (keep the couplings.o dependency on vector.inc, as in the goodhel version)
```

This is essentially your point "1" of your latest email. (And creating the branch valassi_gpucpp_for360 is point "0" of my reply).

I keep this in WIP. But if you find this ok, please by all means go ahead and merge it. **I HAVE NOT TESTED ANY OF GPUCPP_FOR360, SO THIS IS AS-IS**. But you may find it useful for your tests, and in any case this is a merge which needs to happen one way or another.

Hope this helps, Andrea


valassi:
PS Looking at the diffs, this is almost a no-brainer. I think that you should be able to safely merge it. But please have a look and decide.

If we merge it, please do not squash it.

oliviermattelaer:
Yes I think that it makes sense to include this in that branch (which is good since it will trigger the CI and gives us feedback)

PR #141, opened 2024-09-17
This work was to include Dask in the cluster options.

PipGrylls:
The cluster options as given did not work with the system configuration we had. Dask on MPI (and running locally) allowed for many small jobs to work without overwhelming the scheduler with many small tasks.

It was developed but requires domain testing to ensure full compatibility and use.


oliviermattelaer:
Hi,

Thanks, I will test on my cluster (I guess that easybuild can install dask on the cluster) and see if it works.

Now they are couple of file that are change for no reason (but this is just something to clean before doing the merge no issue per say):
1) do not commit input/mg5_configuration.txt
2) do not change .gitignore
3) irrelevant change in misc.py  and  bin/mg5_aMC
4)  file daskmpi_sub_sbatch is clearly not put in the correct place in the repo
Not even sure that such file is actually needed...

Thanks,

Olivier

PR #143, opened 2024-09-25
restructure reweight function to allow Trex plugin

oliviermattelaer:
This is some restructuration of the reweight interface to allow your TRex to be a plugin.
Tell me if you think we need something else.


zeniheisser (approved):
Looks good, no faults seen

PR #144, opened 2024-09-26
installer (remove plugin symlink)

valassi:
Hi @oliviermattelaer this is a WIP PR for the installer

For the moment I only remove the plugin symlink (which I add back in madgraph4gpu)

Later I might add other things

### DIFF
--- a/PLUGIN/CUDACPP_OUTPUT
+++ b/PLUGIN/CUDACPP_OUTPUT
@@ -1,1 +1,0 @@
-../../../epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT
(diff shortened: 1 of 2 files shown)


oliviermattelaer:
I will move forward on the merge of gpucpp with 3.6.0
Such that i can finally release 3.6.0 

So I will already remove the symlink directly in gpucpp.
If you want other change in 3.6.0/gpucpp, it starts to be super super urgent

valassi:
Thanks Olivier. No for the moment I think I am ok with what is there I guess.

Here I added one commit also to disable downloading the hepptoolsinstaller if you use --local. Not strictly needed, if you do not want it then ignore it. I mark this as ready for review.

I then will have something in heptools installer and in madgraph4gpu. But I think not in mg5amcnlo (maybe I forget, then too bad)

oliviermattelaer:
perfect, let's merge this then.

Thanks,

Olivier

PR #148, opened 2024-10-08
Allow patching header/initialise/finalise in driver.f

valassi:
Hi @oliviermattelaer this is one initial commit for allowing patching header/initialise/finalise in driver.f (i.e. to remove eventually patch.P1 etc, as you said many times that you want to do).

Is this going in the right direction?

(And by the way should these things go to 3.x or 3.6.1? I am bit confused, I saw that they diverged so I am not sre which one I should target).

Thanks
Andrea


oliviermattelaer:
Yes this is perfect :-)

For the target, this should be 3.6.1. 
3.x is only updated when a new release is ... release.
Such that user that want to use git will move from one release to the next when they follow 3.x

valassi:
> Yes this is perfect :-)

Very good thanks!

I have not done the changes in cudacpp yet, but this will be useful anyway. I vaguely tested it. Can I merge this already?

> For the target, this should be 3.6.1.

Ok I changed it. Note, I was based on 3.x, so this brings two commits you added in 3.x not yet in 3.6.1. Ok to include those too?

Thanks
Andrea

PR #150, opened 2024-10-09
Add NGOODHEL printout in matrix1.f (move the patch from cudacpp's patch.P1)

valassi:
Hi @oliviermattelaer this is a very simple PR moving one patch from cudacpp patch.P1 to matrix1.f

There is just a debug printout. I hardcoded it. If you prefer, I can add two optional sections, empty in general, filled with additional code for cudacpp.

What do you think?
Thanks
Andrea

### DIFF
--- a/madgraph/iolibs/template_files/matrix_madevent_group_v4.inc
+++ b/madgraph/iolibs/template_files/matrix_madevent_group_v4.inc
@@ -67,7 +67,10 @@ C
 	DATA NB_FAIL /0/
         double precision get_channel_cut
         external get_channel_cut
-	
+C
+        INTEGER NGOODHEL ! -1 if not yet retrieved and printed
+        SAVE NGOODHEL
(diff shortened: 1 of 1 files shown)


oliviermattelaer:
@valassi @Qubitol,
If you still want this, I can merge it in 3.7.0.
I do not think that this is problematic to always have it in.

Qubitol:
I don't have a strong opinion on this, so I'll let you decide what is better.
The alternatives would be to:
- drop these debug statements entirely
- place some `%(cudacpp_extra)` placeholder, and add those statements from the CUDACPP side.

Really, for me every solution works.

valassi:
Hi @oliviermattelaer @Qubitol thanks and sorry for not replying earlier. Yes I'd like to have this merged if possible. It is very useful for debugging in my opinion. Thanks Andrea

PR #151, opened 2024-10-09
Remove valgrind warning in madevent_driver.f template

valassi:
Hi @oliviermattelaer another small PR to slowly move stuff out of cudacpp patchMad.sh

Warning from valgrind is "Conditional jump or move depends on uninitialised value(s)"

Looks ok?

Thanks
Andrea

PR #152, opened 2024-10-09
WIP add optional fortran profiling in various templates

valassi:
Hi @oliviermattelaer this is a WIP PR to add optional fortran profiling in various templates

The idea is to allow the cudacpp patches in https://github.com/madgraph5/madgraph4gpu/pull/962

This is WIP and will stay so for a while. I will mak echep results based on these things but I have no time to clean this up now.

I just wanted to get a heads-up, does this look like the right direction?

Two details
- I have not modified the python yet to define these extra strings as empty by default in mg5amcnlo. And I have not moved the cudacpp patches from patch.P1 to these optional strings.
- I see that three files are .inc which probably allow this mechanism. For dsample.f I guess (I need to check) that this is simply copied. Should this be renamed if it also becomes a template accepting string substitution?

Thanks!
Andrea

PS Ah and this includes internally #151

### DIFF
--- a/Template/LO/Source/dsample.f
+++ b/Template/LO/Source/dsample.f
@@ -175,7 +175,9 @@ c
          if (iter .le. itmax) then
 c            write(*,*) 'iter/ievent/ivec', iter, ievent, ivec
             ievent=ievent+1
+%(start_profiling_x2f)s
             call x_to_f_arg(ndim,ipole,mincfig,maxcfig,ninvar,wgt,x,p)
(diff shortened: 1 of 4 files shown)


valassi:
Note: you had suggested in the past that these pieces of code may become macros controlled by ifdefs. I also tried this out, but in the end here I think your standard method of different code generation is better, more flexible.

A profiling outside cudacpp may always be added at a later time. Fo rthe moment I'd try something like the above.

PR #158, opened 2024-11-21
Signs of life during cross section improvement

zlmarshall:
At the moment, there are cases where cross section improvement can take many hours (we have a case of 5 hours that we're looking at right now). When this sort of thing happens, it is easy for a user to think that the job is stuck in an infinite loop if there is no other sign of life but the CPU churning. This adds a little output every 5th refined job to let folks know that things are proceeding.

### DIFF
--- a/madgraph/madevent/gen_ximprove.py
+++ b/madgraph/madevent/gen_ximprove.py
@@ -1949,6 +1949,11 @@ class gen_ximprove_gridpack(gen_ximprove_v4):
             if j['P_dir'] in done:
                 continue
             done.append(j['P_dir'])
+            # Give a little status. Sometimes these jobs run very long, and having hours without any
+            # console output can be a bit frightening and make users think we are looping.
+            if len(done)%5==0:
+                logger.info(f"Working on job {len(done)} of {len(jobs)}")
+
             # set the working directory path.
             pwd = pjoin(os.getcwd(),j['P_dir']) if self.readonly else pjoin(self.me_dir, 'SubProcesses', j['P_dir'])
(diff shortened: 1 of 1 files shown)


oliviermattelaer:
Great thanks a lot for the idea/patch.

Just by curiosity: Which process/number of events is this happening? (This might be a good candidate for some benchmark)

Will merge this in the branch targetting 3.6.1 relelase.

Thanks,

Olivier

zlmarshall:
Thanks Olivier! The particular process we were dealing with was generation from a gridpack of:

```
generate p p > z{T} z{0} j j QCD=0 QED=4, z > l+ l-
```

Targeting 16,000 events. 51 processes needed refinement, and it was taking ~5h for those 51 processes.

Best,
Zach

oliviermattelaer:
Did you use SDE_strategy=2 for this process (It should be the default one for that process actually)?

zlmarshall:
Looks like we have

```
22:10:06 1	= sde_strategy ! default integration strategy (hep-ph/2021.00773)
```

in the gridpack. I'm not sure what motivated the choice, but if you think it'd be useful I could ping the team and ask them about it.

Best,
Zach

oliviermattelaer:
Yes, I would advise to compare the speed of both sde_strategy=1 and sde_strategy=2, if it follows the result of this paper:
[2102.00773](https://arxiv.org/abs/2102.00773) you can expect to speed-up such computation by large factor (looks like for similar processes, you can be 50 times faster).

In general  I would recomend SDE_strategy=1, but SDE=2 was designed for such type of processes, so here it should be the safest bet.

Cheers,

Olivier

PR #160, opened 2024-12-06
Fix untar parallelization on Mac

Qubitol:
Fix to `restore_data` because it was not working on Mac:
1. `nproc` not present on Mac;
2. `xargs` requires an additional option to work with longer commands.


oliviermattelaer:
Thanks a lot,
-> merging this for 3.6.2 release

Olivier

PR #163, opened 2025-01-09
gridpack multi process execution

roiser:
The run.sh script for the gridpack execution has been augmented with a "-p|--parallel <nprocs>" option which will schedule nprocs processes for the gridpack event generation in parallel. Without the option (or nprocs=1) the gridpack run will fallback into the previous mode of single threaded execution.


roiser:
@oliviermattelaer note this has been tested so far for one process (gg_ttxgg) on CPU, but would like to give an early preview. I'll do more tests in the coming days.

roiser:
Hi @oliviermattelaer , I've added a printout when doing the gridpack generation for the status of waiting, running and done jobs. Featurewise that should be it from my side. Pls have a look,

oliviermattelaer:
So looks like for the moment the code is limited to run one "job" per channel.

Would it make sense to also have a parameter --split_event and automatically split channel if the number of  event for a channel is above that number?

roiser:
Thanks for the comment, actually my initial motivation for this change was to increase the GPU utilisation by coming with multiple parallel processes to be launched. I believe your idea could be beneficial for avoiding the "tail of execution" which I showed in one of our meetings, i.e. one madevent taking most of the time and running for long time while the GPU is not 100% utilised. 

Having said this one could try to first change other parameters e.g. reducing ```max_request_event = 1e12 ``` to avoid this tail.

madgraph/madevent/gen_ximprove.py
```diff
@@ -1945,26 +1953,42 @@ def get_job_for_event(self):
         write_dir = '.' if self.readonly else None  
         self.create_ajob(pjoin(self.me_dir, 'SubProcesses', 'refine.sh'), jobs, write_dir) 
         
+        if self.nprocs > 1:
```
oliviermattelaer, line 1978: The issue is that the use of max_request_event is not in play because this section submit a single job (with the content of info) independently of that value. So we should likely go back to the original function of get_job_for_event (?) or just add a check here and split it accordingly

oliviermattelaer:
@sourcery-ai review

oliviermattelaer:
@sourcery-ai review

PR #165, opened 2025-01-24
Enhance flavour simplified

rikkert-frederix:
Relevant for NLO event generation.

This is a greatly simplified version of the code that was originally developed for biassed event generation based on the flavours present in the process, as used in 2403.14419. Mostly useful for increasing the number of events with b-quarks (before showering) in a multi-jet (merged) sample.

It introduces in new parameter 'Flavour_Bias' to the run card: it requires two values, the absolute value of the PDG code of the flavour to enhance, and the enhancement factor. (Both are integers). For example, setting
 5, 100  = Flavour_Bias
increases the probability of writing an event with an (initial or final) b quark by a factor 100, reducing the corresponding event weight by 100.

### DIFF
--- a/Template/NLO/Source/run.inc
+++ b/Template/NLO/Source/run.inc
@@ -102,3 +102,8 @@ C stuff for alpha running in lepton collisions
       common /to_afromee/ photons_from_lepton
 C  PDFs with beamstrahlung use specific initialisation/evaluation
       common /to_has_bs/ has_bstrahl
+
+
+c for biassed generation based on the particle flavours of the processes
+      integer flavour_bias(0:2)
+      common /c_flavour_bias/ flavour_bias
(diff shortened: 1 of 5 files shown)


oliviermattelaer:
Hi Rik,

For me this can be merged like this. I have just changed the PR to be included in 3.6.2 (is that fine for you?)
 
My only question is if this was possible to do via the custom_fcts function (and if yes would it make sense to provide such functionality via that method --this depend on how much support/visibility you want to provide for such functionality.).
But as you prefer on that.

Cheers,

Olivier

rikkert-frederix:
Hi Olivier,

I thought about putting it in a custom_fcts function, but I couldn't really see a nice way of doing it. I think it's fine the way it is. I'll merge it into 3.6.2

Best,
Rikkert

PR #166, opened 2025-01-24
Faster collect events

rikkert-frederix:
Added an alternative way to collect the events in the various SubProcesses/P*/G* directories (for NLO processes) and write them out in a single event file.

The current method uses a fortran code 'collect_events.f' to randomly read the events from event files, parse them, and then directly write them to the combined event file. The alternative way uses python, reads all the events in memory, randomises them and then writes them to disk. The latter is way faster (in particular for large event files) since all data can be read in one go, and the same for writing the events. However, it requires there to be enough available memory to read all the events. The python code checks the latter explicitly, and if there is not enough memory, it uses the old fortran collect_events instead.

Also, it was not always true to the run_card, param_card, etc. were correctly added to the event file header. This has also been fixed.


madgraph/interface/amcatnlo_run_interface.py
```diff
@@ -3777,37 +3781,281 @@ def reweight_and_collect_events(self, options, mode, nevents, event_norm):
...
+                        mem_info["free"] = int(parts[1])  # In KB
+                    elif parts[0] == "Buffers:":
+                        mem_info["buffers"] = int(parts[1])  # In KB
+                    elif parts[0] == "Cached:":
+                        mem_info["cached"] = int(parts[1])  # In KB
+                # Available memory is free + buffers/cache (and convert to Bytes)
+                available_memory = (mem_info["free"] + mem_info["buffers"] + mem_info["cached"]) * 1024
+                total_memory = mem_info["total"] * 1024
+            return total_memory, available_memory
+
```
marcozaro, line 3824: Wouldn't it be better to have this function in lhe_parser.py?

madgraph/interface/amcatnlo_run_interface.py
```diff
@@ -3777,37 +3781,281 @@ def reweight_and_collect_events(self, options, mode, nevents, event_norm):
...
+        def read_events_from_file(file_path):
```
marcozaro, line 3825: Wouldn't it be better to have this function in lhe_parser.py?

madgraph/interface/amcatnlo_run_interface.py
```diff
@@ -3777,37 +3781,281 @@ def reweight_and_collect_events(self, options, mode, nevents, event_norm):
...
+                        event_lines = [line]
+                    elif "</event" in line:
+                        event_lines.append(line)
+                        events.append("".join(event_lines))
+                        inside_event = False
+                    elif inside_event:
+                        event_lines.append(line)
+            return events
+        
+        def replace_weight_in_event(event, main_weight,iproc):
```
marcozaro, line 3843: Wouldn't it be better to have this function in lhe_parser.py?

madgraph/interface/amcatnlo_run_interface.py
```diff
@@ -3777,37 +3781,281 @@ def reweight_and_collect_events(self, options, mode, nevents, event_norm):
...
+                    in_rwgt_section = False
+                elif in_rwgt_section and "<wgt" in line:
+                    # Extract and rescale the weight YYY inside each <wgt id='XX'> YYY </wgt>
+                    lines[i] = re.sub(r"(<wgt id='\d+'>)(\s*-?\d*(\.\d*)?[eE][-+]?\d+)(\s*</wgt>)",
+                                      lambda match: f"{match.group(1)} {float(match.group(2)) * scale_factor:.4e}{match.group(4)}",
+                                      line)
+            
+            return "\n".join(lines) + "\n"
+        
+        def get_header_footer(file_path):
```
marcozaro, line 3873: Wouldn't it be better to have this function in lhe_parser.py?

madgraph/interface/amcatnlo_run_interface.py
```diff
@@ -3777,37 +3781,281 @@ def reweight_and_collect_events(self, options, mode, nevents, event_norm):
...
+            return header, footer
+        
+        def write_randomized_events(output_file, events, header, footer):
+            """Writes header, randomized events, and footer to the output file."""
+            with open(output_file, 'w') as f:
+                f.writelines(header)
+                f.writelines(events)
+                f.writelines(footer)
+
+        def add_banner_to_header(header,tot_xsec,tot_xerrABS,main_weight,iprocs):
```
marcozaro, line 3894: Wouldn't it be better to have this function in lhe_parser.py?

madgraph/interface/amcatnlo_run_interface.py
```diff
@@ -3777,37 +3781,281 @@ def reweight_and_collect_events(self, options, mode, nevents, event_norm):
...
+            with open(file_path, 'r') as f:
+                for line in f:
+                    if "<event" in line:
+                        inside_events_section = True
+                        break
+                    header.append(line)
+            footer = "</LesHouchesEvents>"
+            return header, footer
+        
+        def write_randomized_events(output_file, events, header, footer):
```
marcozaro, line 3887: Wouldn't it be better to have this function in lhe_parser.py?

marcozaro (commented):
Hi Rik,
good, I have a first round of 'bigger' changes. I will have a look at the fine details later.
In general, why not using directly the lhe_parser.py module?
There is MultiEventFile in there which should be handy...

Thanks!

Best,

Marco

rikkert-frederix:
Hi Marco,

Thanks for your suggestion of moving many functions to the lhe_parser.py. I agree that that would be a much better place. I'll have a look on how to do that in practice.

Best,
Rikkert

PR #168, opened 2025-02-11
Setup an interpolation grid for alphaS. 

rikkert-frederix:
Relevant for FxFx, since the Sudakov is computed numerically, which requires MANY alphaS evaluations. Grid is faster; reduces computation time by a little bit. (Only affects the build-in PDF sets: when using LHAPDF the computation of alphaS is done by LHAPDF).

### DIFF
--- a/Template/NLO/Source/alfas_functions.f
+++ b/Template/NLO/Source/alfas_functions.f
@@ -71,8 +71,8 @@ ccc
 
       DOUBLE PRECISION FUNCTION ALPHAS(Q)
 
-      DOUBLE PRECISION Q,ALPHAS_not_timed
-      external ALPHAS_not_timed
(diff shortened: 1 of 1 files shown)


marcozaro:
Ciao Rikkert,
thanks a lot. I guess you checked that the result is compatible with the 'standard' evaluation of alphas.
The code changes look fine. Would it be an option to keep both the exact and the interpolated alphas, and to switch among them via a (hidden) run-card variable?

Cheers,

Marco

rikkert-frederix:
Hi Marco,

I don't think we need a parameter in the run_card. Note that also the original implementation was not 'exact', but rather obtained through a zero-finding algorithm that was executed to 5x10^-4 precision. 

Best,
Rikkert

PR #170, opened 2025-02-26
3.6.2 color symmetric

oliviermattelaer:
This PR is implementing something that I though was merged two years ago.

The main idea is to avoid to do the computation of the Full color matrix and only compute the upper diagonal part (including the diagonal obviously). This speed-up g g > 5g by 40% and tt~3g by ~25%

To have faster compilation I have 
- restore the color matrix as an integer matrix rather than double precision.
- change the storing format of the color matrix from a 2D array to a 1D array.

This PR also fix the test-suite to have it all green.

## Summary by Sourcery

Optimize the color matrix calculation by computing only the upper diagonal part, which improves performance for processes with g g > 5g by 40% and tt~3g by ~25%. Additionally, improve compilation speed by restoring the color matrix as an integer matrix and changing the storing format to a 1D array.

Enhancements:
- Optimize the color matrix calculation to compute only the upper diagonal part, improving performance.
- Restore the color matrix as an integer matrix rather than double precision to improve compilation speed.
- Change the storing format of the color matrix from a 2D array to a 1D array for faster compilation.
- Update the CI workflow to use ubuntu-22.04

Tests:
- Fix the test-suite to have it all green.


oliviermattelaer:
To update after today discussion, passing on "parameter" is actually non trivial since it force the color matrix to be defined in a single line which can be problematic given the size of that matrix for the case where it can be relevant.

Also fortran does not have the "need" to know that memory is constant since you are much more limited in term of pointer, fortran can determine at compile time that the CF is not changed when doing the color computation, (while C++ can not determine/assume that and thererfore need the const in order to apply the same optimization as fortran)

oliviermattelaer:
@sourcery-ai review

oliviermattelaer:
@sourcery-ai review

marcozaro:
Ciao Olivier,
sorry for the delay. Excellent job!
I checked an NLO process. Born and reals have the color matrix in the new format. The color-linked born (b_sf_xxx.f) do not.
Did you choose on purpose not to change these? 
In case, we can take a look together, as these are quite tricky (you have two different color bases to interfere, and you may end up with non-square color matrices)

Let me know if you want.
Thanks again,

Marco

rikkert-frederix:
Hi Olivier,

I finally had some time to look at this, and it looks all good. No reservations from my side regarding pulling this into the main code.

best,
Rikkert

oliviermattelaer:
Thanks to you both,

@marcozaro, I can switch those to integer format, but not to upper diagonal mode (which will not make sense here, I guess).

Cheers,

Olivier

marcozaro:
OK, thanks, if you could, that will be great!

Thanks!

Marco

oliviermattelaer:
@marcozaro 
Done and ... I did a typo, and the test suite detected it (which is nice).

Except if you object, I guess this is good to go.

Thanks,

Olivier

marcozaro:
ok! Thanks!

PR #179, opened 2025-05-07
Second option for externals include path

zlmarshall:
For many installations, the headers are in an include directory, but then in a sub-directory that has the name of the package. This just provides a second option for looking for headers (it just so happens that this is the structure used by genser on cvmfs as well)

There are a bunch of ways to do this, and this is only one, so no hard feelings if you'd prefer a different implementation!

Cheers,
Zach

## Summary by Sourcery

Enhance header include path detection for external libraries by adding a second search option in a package-specific subdirectory

Bug Fixes:
- Improve error logging to show both searched include paths when header detection fails

Enhancements:
- Modify include path detection to check for headers in a secondary location within an 'include' directory


oliviermattelaer:
Sounds perfect to me. 
I think it does make sense to consider this as a "bug" and therefore to also include in the 3.5.X LTS branch (but probably not for the 2.9.x LTS). I will cherry pick it into that branch afterwards but directly merge it here within the 3.6.3 release.

Thanks a lot,

Olivier

zlmarshall:
Thank you for your help!

PR #181, opened 2025-05-15
3 x index out of range

richardphysics:
Fixing "list index out-of-range" bug that appeared between 3.6.1 and 3.62

## Summary by Sourcery

Fix list index out-of-range errors and improve model import, phase‐space constraints, and color algebra while introducing new validation checks and HEFT acceptance testing.

New Features:
- Add model consistency checks (check_model_all and check_model_aS) to enforce α_s assignment in SMINPUTS#3.

Bug Fixes:
- Constrain goldstone–vector merging to spin-3 candidates to prevent list index errors.
- Refine goldstone interaction merging to avoid duplicate vertices.
- Fix histogram parsing by removing erroneous weight header boundary checks.

Enhancements:
- Introduce dsqrt_shatmax parameter and enforce ŝ limits in phase-space generation, CMS sampling, and cuts.
- Extend ALOHA CombineAmpS to support larger wave function sizes.
- Incorporate denominator scaling and symmetric indexing in color amplitude templates.
- Enhance FKS export include path search with secondary directory fallback and clearer error messages.
- Update vertex color reordering logic to handle deep-copy anomalies and proper ColorString conversion.
- Use maximum instead of minimum denominator in export_v4 for consistent color matrix data.

CI:
- Add acceptancetest_heft job in GitHub Actions for HEFT generation.

Tests:
- Add test_generation_heft acceptance test to catch HEFT exporter crashes.

### DIFF
--- a/madgraph/interface/common_run_interface.py
+++ b/madgraph/interface/common_run_interface.py
@@ -1995,18 +1995,28 @@ class CommonRunCmd(HelpToCmd, CheckValidForCmd, cmd.Cmd):
                                        self.run_card['event_norm'] in ['unity']:
                 all_cross= [cross/nb_event for cross in all_cross]
                 
-            sys_obj = systematics.call_systematics([input, None] + opts, 
+            # RR 2025 0515
+            # 'old_run_mode' only defined to bypass crash.
+            # if exists, no need to re-re-run print
+            if 'old_run_mode' in locals():
+                self.update_status('Skipping (re)run of call_systematics', level='parton', makehtml=False)
+                self.update_status('Skipping (re)run of print_cross_sections', level='parton', makehtml=False)
+                self.update_status('Skipping (re)concatenation', level='parton', makehtml=False)
+            else:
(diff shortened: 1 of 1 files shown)


oliviermattelaer:
Would not make more sense to "just return after the try/except part of the code?"

Olivier

richardphysics:
> Would not make more sense to "just return after the try/except part of the code?"
> 
> Olivier

Hi Olivier,
I looked into that. It does work. However then the tmp_N_unweighted files are not removed. These are removed at the very end of the function by the "for i in range(nb_submit):" loop.

oliviermattelaer:
What would you think to do it like this then?

richardphysics:
hi olivier, this looks good! I like moving a copy of  the "remove routine" higher into the exception clause. I also had no idea you could insert an exception *into* the argument of a function. cool!

fyi, the EVA@NLP  update has my "out-of-range" solution, so when merging this will need to be updated.

PR #185, opened 2025-07-21
Fixes two bugs in FxFx

rikkert-frederix:
This fix corrects two bugs in FxFx merging, introduced in version 3.3.1.

One is a "coding bug", in driver_mintMC.f the FxFx terms relevant to the n+1-body contributions were computed before the prefactors for these contributions was set, resulting in overwriting of the FxFx contributions.

The second is the application of the generation-level cuts. From versions 3.3.2 these were applied exclusively on the clustering scales, ignoring any dependence on DeltaR. This is incorrect at large pTs, were these should be taken into account.

This should also be fixed in the LTS version.

## Summary by Sourcery

Fix two bugs in FxFx merging: ensure n+1-body prefactors are computed before scale resets to prevent overwriting FxFx contributions, and apply generation-level cuts correctly by introducing a pt-dependent jet threshold and a need_matching_cuts flag to account for DeltaR at high pT

Bug Fixes:
- Compute n+1-body prefactors before resetting FxFx scales in driver_mintMC to preserve FxFx terms
- Introduce need_matching_cuts and replace fixed sycut with pt-dependent threshold in passcuts_fxfx to enforce generation-level cuts including DeltaR

### DIFF
--- a/Template/NLO/SubProcesses/cluster.f
+++ b/Template/NLO/SubProcesses/cluster.f
@@ -475,7 +475,7 @@ c     the renormalisation scale.
 c Determine which particles need clustering and which do not
       call matching_particles(next,nbr,ipdg,cluster_pdg,cluster_ij,iord
      $     ,need_matching)
-
+      
       nqcdrenscale=0       ! number of alpha-s needing reweighting
       nqcdrenscalecentral=0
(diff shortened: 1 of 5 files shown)


oliviermattelaer (approved):
Perfect, nothing to comment on. Let's merge this

PR #189, opened 2025-08-13
Fix borntilde

oliviermattelaer:
Hi Rik,

So here is the PR to fix the amount of failed soft and collinear test in multiple process.
As you were identifying this is indeed related to the computation of the color-matrix.

However, the argument that the element on the left and right are not identical is not enough to break the implementation.
(which is good news since this means that the implementation is fine for interference computation).

In order to break what I did, you also need to not sum symmetrically over all helicity, which I think is the case due the condition NHEL(GLU_IJ,IHEL).EQ.NHEL(GLU_IJ,1)
(which likely explains why MadSpin was not impacted)

I will continue to investigate this further for the density branch, but anyway this should be merge fast.

Olivier

## Summary by Sourcery

Fix incorrect Born-tilde color correlator computation by summing color factor contributions symmetrically over helicities and correcting the normalization divisor.

Bug Fixes:
- Include both interference terms (JAMPH(2,J,M)*DCONJG(JAMPH(1,I,N)) and JAMPH(2,I,M)*DCONJG(JAMPH(1,J,N))) in the BORNTILDE and ANS calculations.
- Adjust normalization to divide by 2*DENOM instead of DENOM for the Born-tilde contributions.

Tests:
- Update generated test input files to reflect the corrected color-matrix calculations.

### DIFF
--- a/madgraph/iolibs/template_files/born_cnt_splitorders_fks.inc
+++ b/madgraph/iolibs/template_files/born_cnt_splitorders_fks.inc
@@ -351,15 +351,14 @@ c          Jamp2(i,J)=Jamp2(i,J)+Jamp(i,J)*dconjg(Jamp(i,J))
         ZTEMP = (0.D0,0.D0)
         DO J = I, NCOLOR
           CF_INDEX = CF_INDEX + 1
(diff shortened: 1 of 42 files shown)


rikkert-frederix (approved):
Thanks, Olivier, it looks good!

PR #191, opened 2025-08-19
Changes to implement CUDACPP patches in upstream

Qubitol:
Implement the changes originally imported via `patchMad.sh` from the CUDACPP plugin.
They are protected through `#ifdef`, and makefile has been updated to accept preprocessor macros with the `-cpp` flag.

See also related [madgraph4gpu#1040](https://github.com/madgraph5/madgraph4gpu/pull/1040).

## Summary by Sourcery

Implement upstream support for the CUDACPP plugin in the Fortran driver and enable preprocessing in the build system.

New Features:
- Add MG5AMC_MEEXPORTER_CUDACPP guard in madevent_driver.f to initialize and finalize the CUDA-accelerated Fortran-CPP bridge (FBRIDGE)
- Allow configuring FBRIDGE mode and vector sizes at runtime via CUDACPP_RUNTIME_FBRIDGEMODE and CUDACPP_RUNTIME_VECSIZEUSED environment variables
- Output MERATIOS performance metrics when running in mixed debug or quiet modes

Enhancements:
- Enable Fortran preprocessing by adding the -cpp flag to FC invocations in Template/LO/SubProcesses makefile

### DIFF
--- a/madgraph/iolibs/template_files/madevent_driver.f
+++ b/madgraph/iolibs/template_files/madevent_driver.f
@@ -93,6 +93,58 @@ C  BEGIN CODE
 C----- 
       call cpu_time(t_before)
       CUMULATED_TIMING = t_before
+
+#ifdef _OPENMP
+      CALL OMPNUMTHREADS_NOT_SET_MEANS_ONE_THREAD()
+#endif
+
+#ifdef MG5AMC_MEEXPORTER_CUDACPP
+      CALL COUNTERS_INITIALISE()
+      fbridge_mode = 1 ! CppOnly=1, default for CUDACPP
+      env_name = 'CUDACPP_RUNTIME_FBRIDGEMODE'
+      call get_environment_variable(env_name, env_value, env_length, env_status)
+      if( env_status.eq.0 ) then
+        write(*,*) 'Found environment variable "', trim(env_name), '" with value "', trim(env_value), '"'
+        read(env_value,'(I255)') FBRIDGE_MODE ! see https://gcc.gnu.org/onlinedocs/gfortran/ICHAR.html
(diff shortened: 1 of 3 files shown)


Qubitol:
Remove `#ifdef`s and the contents of the patches themselves, and use string placeholders: `CUDACPP_EXTRA_HEADER`, `CUDACPP_EXTRA_INITIALISE`, `CUDACPP_EXTRA_FINALISE`, so that the additions are done directly by the CUDACPP plugin.

@sourcery-ai review

madgraph/iolibs/template_files/madevent_driver.f
```diff
@@ -76,13 +76,14 @@ Program DRIVER
 
       include 'coupl.inc' ! needs VECSIZE_MEMMAX (defined in vector.inc)
       INTEGER VECSIZE_USED
-      DATA VECSIZE_USED/VECSIZE_MEMMAX/ ! can be changed at runtime
+%(CUDACPP_EXTRA_HEADER)s
 %(DRIVER_EXTRA_HEADER)s
```
oliviermattelaer, line 80:
```suggestion
%(DRIVER_EXTRA_HEADER)s
```
Do we need (i.e. is it an interest) to have an additional flag here?
I mean that in principle adding in front of the value "driver_extra_header" should be enough to edit the file as we want.

I'm fine both way, but I  see two advantages:
 - less keyword
 - potentially less issue to handle old version

oliviermattelaer (approved):
They are still one point that annoys me a bit (sorry for being annoying here).
But this is detail and I'm ok to merge like this if you prefer it like this.
So I already approve it.

Olivier

PR #193, opened 2025-09-18
change nlo rule to be more nlo friendly

oliviermattelaer:
Hi,

This is mostly to open the discussion and show how we can improve the situation.
This was trigger by a report of mawatari:
That the syntax:
import model DMsimp_s_spin1; generate p p > xd xd~ j [QCD]
crash in 3.x while it was working fine in 2.x

The reason is the automatic restriction of coupling square which are systematically set to zero.

The patch that I propose here, relax this condition by handling coupling order according to their hierarchy.
In top of that if the nlo_mixed_expansion is False then I relax the condition even more (but likely not needed in practise).

Cheers,

Olivier

## Summary by Sourcery

Relax the automatic zeroing of non-QED/QCD coupling orders in NLO process generation by assigning orders based on the model's order_hierarchy and expansion_order to prevent crashes.

Bug Fixes:
- Prevent crashes when generating NLO processes with non-QED/QCD couplings by avoiding indiscriminate zeroing of coupling squares

Enhancements:
- Assign non-QED/QCD coupling orders according to the model's order_hierarchy instead of setting them to zero
- Further relax coupling order restrictions when nlo_mixed_expansion is False by incorporating the model's expansion_order for eligible couplings

### DIFF
--- a/madgraph/interface/amcatnlo_interface.py
+++ b/madgraph/interface/amcatnlo_interface.py
@@ -563,7 +563,12 @@ Please also cite ref. 'arXiv:1804.10017' when using results from this code.
                 # set all the other coupling to zero
                 for o in myprocdef['model'].get_coupling_orders():
                     if o not in ['QED', 'QCD']:
-                        orders[o] = 0
+                        if self._curr_model.get('order_hierarchy')[o] == self._curr_model.get('order_hierarchy')['QCD']:
+                            orders[o] = 2*qcd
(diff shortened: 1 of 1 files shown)


rikkert-frederix:
Hi Olivier,
I think this fix is fine. It only affects BSM couplings, and it should be safe to set them the way you do.
Rikkert

rikkert-frederix (approved):

PR #196, opened 2025-10-01
remove withespace

oliviermattelaer:
This (should be) the same as PR: #192 
But with running some command to avoid all the spaces. which make the review easier/nicer.

Can you check that everything is fine with this version?

Script that I did run to get this:
```
git switch 3.7.0
git checkout -b UPC_NLO_no_space
git merge --no-commit UPC_NLO
git reset
git diff -U0 -w --no-color | git apply --cached --ignore-whitespace --unidiff-zero -
git checkout .
git add Template/LO/Source/PDF/gammaUPC/elliptic_integral.f90 Template/LO/Source/PDF/gammaUPC/fmin_module.f90 Template/LO/Source/PDF/gammaUPC/photoabsorption.f90 Template/NLO/Source/PDF/gammaUPC/ElasticPhotonPhotonFlux.f90 Template/NLO/Source/PDF/gammaUPC/OpticalGlauber_Geometry.f90 Template/NLO/Source/PDF/gammaUPC/elliptic_integral.f90 Template/NLO/Source/PDF/gammaUPC/fmin_module.f90 Template/NLO/Source/PDF/gammaUPC/gammaUPC_dummy.f Template/NLO/Source/PDF/gammaUPC/interpolation.f90 Template/NLO/Source/PDF/gammaUPC/makefile Template/NLO/Source/PDF/gammaUPC/makefile_dummy Template/NLO/Source/PDF/gammaUPC/nielsen_generalized_polylog.f90 Template/NLO/Source/PDF/gammaUPC/nintlib.f90 Template/NLO/Source/PDF/gammaUPC/photoabsorption.f90 Template/NLO/Source/PDF/gammaUPC/photonpdfsquare.f Template/NLO/Source/PDF/gammaUPC/run90.inc Template/NLO/Source/PDF/gammaUPC/tbessj.f90 Template/NLO/Source/PDF/gammaUPC/tbessk.f90 Template/NLO/Source/PDF/gammaUPC/test.f90
git commit -a -m "remove_space"
```


madgraph/core/diagram_generation.py
```diff
@@ -1730,11 +1730,25 @@ def generate_multi_amplitudes(cls,process_definition,
         except KeyError:
             fstags = []
 
+        try:
+            istags = [leg['is_tagged'] for leg in process_definition['legs'] \
+                 if 'is_tagged' in leg.keys() and leg['state'] == False]
+
+        except KeyError:
+            istags = []
+
```
oliviermattelaer, line 1739: This sounds a quite dangerous error catching, maybe this one is better (depending which error you want to catch)
```suggestion
        try:
            process_definition['legs']
        except KeyError:
            istags = [] 
        else:
            istags = [leg['is_tagged'] for leg in process_definition['legs'] \
                 if 'is_tagged' in leg.keys() and leg['state'] == False]
```

madgraph/core/diagram_generation.py
```diff
@@ -1730,11 +1730,25 @@ def generate_multi_amplitudes(cls,process_definition,
...
         # Generate all combinations for the initial state
         for prod in itertools.product(*isids):
+            if any(istags):
+                if not all(istags):
+                    logger.error("Tagging only one initial-state particle is not allowed")
+                    exit()
```
oliviermattelaer: exit quit the code inconditionally which is not nice in interactive mode (and bypass the creation of debug file)

```suggestion
                    raise MadGraph5Error("Tagging only one initial-state particle is not allowed")
```

madgraph/fks/fks_common.py
```diff
@@ -287,10 +287,19 @@ def find_splittings(leg, model, dict, pert='QCD', include_init_leptons=True): #t
                         # check if the leg is tagged, that 
                         # the same particles appear also in the two daughters
                         if 'is_tagged' in leg.keys() and leg['is_tagged'] and \
+                           leg['state'] and \
```
oliviermattelaer, line 290: maybe add a comment here why you add a condition

madgraph/iolibs/export_fks.py
```diff
@@ -1158,7 +1158,8 @@ def write_a0gmuconv_file(self, writer, matrix_element):
         bornproc = matrix_element.born_me['processes'][0]
         startfromalpha0 = False
         if any([l['is_tagged'] and l['id'] == 22 for l in bornproc['legs']]):
-            if 'loop_qcd_qed_sm_a0' in bornproc['model'].get('modelpath'):
+            if 'loop_qcd_qed_sm_a0' in bornproc['model'].get('modelpath') \
+               or 'loop_qcd_qed_MSbar' in bornproc['model'].get('modelpath'):
```
oliviermattelaer: This mechanism is ugly.
We should change it to either move such mechanism definition to the UFO itself or at least via a list in input/model_a0_scheme.dat

Also is it really what you want (sounds weird to have a MSbar model activating this)

madgraph/various/banner.py
```diff
@@ -1190,7 +1190,7 @@ def __setitem__(self, name, value, change_userdefine=False,raiseerror=False):
                     
                 if not new_values:
 
-                    text= "value '%s' for entry '%s' is not valid.  Preserving previous value: '%s'.\n" \
+                    text= "value '%s' for entry '%s' is not valid.  1 Preserving previous value: '%s'.\n" \
```
oliviermattelaer:
```suggestion
                    text= "value '%s' for entry '%s' is not valid. Preserving previous value: '%s'.\n" \
```

madgraph/various/banner.py
```diff
@@ -1296,7 +1296,7 @@ def __setitem__(self, name, value, change_userdefine=False,raiseerror=False):
                     
             if not valid:
                 # act if not valid:
-                text = "value '%s' for entry '%s' is not valid.  Preserving previous value: '%s'.\n" \
+                text = "value '%s' for entry '%s' is not valid.  2 Preserving previous value: '%s'.\n" \
```
oliviermattelaer:
```suggestion
                text = "value '%s' for entry '%s' is not valid. Preserving previous value: '%s'.\n" \
```

madgraph/various/banner.py
```diff
@@ -5517,6 +5517,25 @@ def warn_fifo(input):
...
+  %(nb_proton1)s    = nb_proton1 # number of proton for the first beam
+  %(nb_neutron1)s    = nb_neutron1 # number of neutron for the first beam
+# Note that seting differently the two beams only work if you use
+# group_subprocess=False when generating your matrix-element
+  %(nb_proton2)s    = nb_proton2 # number of proton for the second beam
+  %(nb_neutron2)s    = nb_neutron2 # number of neutron for the second beam"""
+template_off = "# To see heavy ion options: type \"update ion_pdf\""
+
+heavy_ion_block_nlo = RunBlock('ion_pdf', template_on=template_on, template_off=template_off)
+
```
oliviermattelaer: Do you (really) need a different block compare to the LO one?
Would be nicer to have the exact same block. Additionally do you not need the mass of the particles?
```suggestion

```

madgraph/various/banner.py
```diff
@@ -5532,7 +5551,7 @@ class RunCardNLO(RunCard):
      
     LO = False
     
-    blocks = [running_block_nlo]
+    blocks = [heavy_ion_block_nlo, running_block_nlo]
```
oliviermattelaer:
```suggestion
    blocks = [heavy_ion_block, running_block_nlo]
```

madgraph/various/banner.py
```diff
@@ -5681,11 +5708,21 @@ def check_validity(self):
...
+            if self['nb_proton2'] !=1 or self['nb_neutron2'] !=0:
+                raise InvalidRunCard( "Heavy ion mode is only supported for lpp2=1/2")
+
         # for lepton-lepton collisions, ignore 'pdlabel' and 'lhaid'
         if abs(self['lpp1'])!=1 or abs(self['lpp2'])!=1:
             if self['lpp1'] == 1 or self['lpp2']==1:
                 raise InvalidRunCard('Process like Deep Inelastic scattering not supported at NLO accuracy.')
 
+            if not self['lpp1'] == self['lpp2'] == 2:
+
```
oliviermattelaer: Not sure to understand such if ... 
Can you check what is the point and if the space handling was not "wrong" here due to the automatic removal of space issue?

Template/NLO/SubProcesses/makefile_fks_dir
```diff
@@ -22,7 +22,7 @@ NLOLIBS=-L$(LIBDIR) $(libmadloop) $(libcuttools) $(libOLP)
 # Uncomment the line below to compile with MadLoop without setting the environnement 
 # variables 'by hand'
 # NLOLIBS= -L. -lMadLoop -L$(LIBDIR) -lcts
-LINKLIBS = -L$(LIBDIR) -ldhelas -lgeneric -lmodel -lpdf -lcernlib $(llhapdf)
+LINKLIBS = -L$(LIBDIR) -ldhelas -lgeneric -lmodel -lpdf -lgammaUPC -lcernlib $(llhapdf)
```
oliviermattelaer, line 25: Would it be possible to not compile gammaUPC if those PDF are not used?
That would mainly means to link here only against lpdf
and that the PDF makefile link against it according to the value of the run_card (which should already be the case for the other PDF --at least at LO--)

oliviermattelaer (commented):
One global comment, is about the gammaUPC directory.
Would it be possible to have it in Common/Source/PDF
rather than having two copies?
Or if that's an issue to have symlink for identical files?

This PR seems to have broken one if statement "tab adjustment".
Which needs to be fixed (sorry this is my fault technically)

Cheers,

Olivier

madgraph/iolibs/export_fks.py
lsimon125: I agree, this mechanism is a bit ugly, but it was already introduced for tagged final state photons and we basically copied it. I think a cleaner way for this might require more deep changes in the code.

lsimon125: It's correct that the MSbar model activates this. This particular model was used for some previous studies in electron-positron collisions and older gamma-gamma studies.

madgraph/various/banner.py
lsimon125: Here was an issue that remove the indentation when removing the whitespaces. I have fixed it.

Template/NLO/SubProcesses/makefile_fks_dir
lsimon125: We use the same mechanism here as for LO. In case gamma-UPC is not use, a simple dummy library is compiled, but not the whole gamma-UPC library.

I also followed your recommendation and moved gamma-UPC to the common 'Template' directory instead of having two copies.

lsimon125:
I have corrected the bugs introduced by the whitespace changes and addressed all comments through updates to the code. Please let me know if everything looks good now.

Cheers,
Lukas

madgraph/fks/fks_common.py
lsimon125: done

madgraph/iolibs/export_fks.py
oliviermattelaer: Ok we should have a discussion on such point. I will not block the PR for that,
but we need to clean such stuff, it is not good practise to hardcode behavior based on the model name
(which means changing the model name for test purpose will change the behavior)

oliviermattelaer: issue created

madgraph/core/diagram_generation.py
oliviermattelaer: Do you agree with this suggestion?
If you do then I can do the merge of this branch

lsimon125: Sorry, I forgot to check this comment. To be honest, I cannot remember the KeyError I was trying to catch. I think it was related to the key 'is_tagged', but I need to check this again.

oliviermattelaer: This is exactly the reason why those except need to be "clear" otherwise it is difficult to maintain.

A potential missing "is_tagged" should be handled via the 
```
if 'is_tagged' in leg.keys()
```
so in that case maybe even better to do
```suggestion
        istags = [leg['is_tagged'] for leg in process_definition['legs'] \
                 if 'is_tagged' in leg.keys() and leg['state'] == False]
```

madgraph/iolibs/export_fks.py
oliviermattelaer: So from the issue, we did agree to treat that differently. I have update those lines (if this is indeed the same one) so one just need to update the model (and follow what i did in the other branch)

lsimon125: I will tag @huashengshao since he created the UFO to adapt the changes discussed in #202.

madgraph/core/diagram_generation.py
lsimon125: I agree with your latest solution, it works well, and I don’t see any need for the `try-except` construction. I probably just copied it from the previous lines, where the same pattern is used for tagged final-state particles. Maybe @marcozaro remembers why it was implemented that way; I don’t see the need for it there either.

PR #199, opened 2025-10-19
fix: hepmc2 extrapaths for system installed pythia

APN-Pucky:
Running on my computer:
```bash
$ pythia8-config --hepmc2
-lHepMC
```
results in the `extrapath` append code failing with an `IndexError`.

## Summary by Sourcery

Bug Fixes:
- Iterate over hepmc2 output tokens to extract '-L' library paths instead of using a fixed index, avoiding IndexError when no '-L' flag is present

### DIFF
--- a/UpdateNotes.txt
+++ b/UpdateNotes.txt
@@ -4,6 +4,11 @@ ANNOUNCEMENT:
    2.9.X version as a LONG TERM STABLE has now an end of life for bug fixing/support in december 2025.
    A new LTS (based on current 3.5.X) is starting now and will act as stable release for the coming years.
 
+3.6.5 (17/10/25):
+     MZ: Ensure that the code is crashing at NLO when the analysis contains a syntax error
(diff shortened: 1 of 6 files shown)


oliviermattelaer:
Excellent, thanks a lot,

Olivier

oliviermattelaer:
Looks like the CI crashed after having merged this ... I'm reverting this for the moment to check if this is related to this...

APN-Pucky:
> Looks like the CI crashed after having merged this ... I'm reverting this for the moment to check if this is related to this...

I ran my changes and they passed for my case (where they failed previously with an IndexError). I am not sure about the other commits you added to this MR tho. It seems like the CI still fails with my 2 commits reverted...

oliviermattelaer:
Correct, looks like 3.7.0 already had CI not passing, but I failed to react/notice them.
I have ping the responsible, and will restore those commits.

Sorry for the noise

Olivier

PR #201, opened 2025-10-27
Add the density mode to madspin_density branch

valentindurupt:
Merging the branch madspin_valentin into madspin_density.

This mode adds an option to the reweight mode that allows to compute density matrices event by event and written in the LHE file. The user's input are given in the reweight card.


madgraph/iolibs/template_files/matrix_standalone_v4.inc
```diff
@@ -461,10 +460,30 @@ C
...
+      DOUBLE PRECISION IMAG_RHO(99)
+      INTEGER I
+
+      DO I = 1, 99
+        REAL_RHO(I) = DREAL(RHO(I))
+        IMAG_RHO(I) = DIMAG(RHO(I))
+      ENDDO
+
+      RETURN
+      END
```
oliviermattelaer: I'm confused about this function. How do you use it?
From the definition I would expect that you would have some line in python like
real, imag = m0_transfer_density(rho)

but in that case the "rho" object is already within python and this function does not do anything ...

madgraph/various/Density_functions.py
```diff
@@ -0,0 +1,1155 @@
...
+def Fidelity_distance(Matrix1: list[complex, complex], Matrix2: list[complex, complex]) -> float:
+    """
+       Input: Matrix1, Matrix2 -> 2 square matrices
+       Output: fidelity distance between the two matrices
+    """
+    fidelity = Fidelity(Matrix1, Matrix2)
+    return np.sqrt(1 - fidelity**2)
+
+
+#########################END QI OBSERVABLES ############################
```
oliviermattelaer: Here my only comment is why not using object oriented programming for such function.
Most of them are taking rho / Matrix entry.
So defining them within the density class would likely make it easier to find/handle in the future no ?

madgraph/interface/reweight_interface.py
```diff
@@ -1598,14 +1662,526 @@ def calculate_matrix_element(self, event, hypp_id, scale2=0):
...
+                    for j in range(len(pdg)):
+                        if pdg[new_order_pos[j]] == pos_aux[i] and is_particle_taken_new[j] == 0:
+                            pos_new[compteur_pos_new] = new_order_pos[j] + 1 #+1 because fortran format
+                            is_particle_taken_new[j] = 1
+                            compteur_pos_new += 1
+                            break #needed to put only one particle for each pos_aux[i]
+
+                pos_corrected = pos_new
+
+#######################END PARTICLE CHOICE BLOCK#############################
```
oliviermattelaer: This would need to be refactorised in a suite of smaller function.

Would be also nice if we could avoid as many as possible the 
"if self.flag_density_matrix:"
by creating more function within the standard class (potentially not doing anything in the standard class).
such that most of those "if" are hidden within the class definition.

oliviermattelaer (changes requested):
I think it will be worth the time to refactorise a bit the code to use more object oriented style of programming and avoid long function (which are difficult to read).

Do hesitate to come to my office for help on that

madgraph/iolibs/template_files/matrix_standalone_v4.inc
valentindurupt: The function has been removed.

madgraph/various/Density_functions.py
valentindurupt: I have rewritten in an object oriented way. Some auxiliary functions are still kept outside the classes DensityMatrixObservables / DensityMatrixObservables22 / DensityMatrixObservables23 / DensityMatrixObservable33.

madgraph/interface/reweight_interface.py
valentindurupt: Done. There are still a couple of "if self.flag_density_matrix:" but I am not certain it is worth to get rid of them.

PR #205, opened 2025-11-03
Fix quotes and permissions in PYTHIA8 script

APN-Pucky:
I got this error:
```
chmod: changing permissions of '/usr/bin/pythia8-config': Operation not permitted
make: *** No rule to make target '/usr/lib/libpythia8.a', needed by 'Pythia83'.  Stop.
```
which is already executable and owned by root.

This changes the code to only call `chmod +x` if it is not already executable. Also `$PYTHIA8LOCATION` could have spaces => quote it.

## Summary by Sourcery

Fix quoting and conditional permission changes in the PYTHIA8 setup script

Bug Fixes:
- Quote $PYTHIA8LOCATION path to support locations containing spaces
- Only invoke chmod +x when the target file is not already executable


oliviermattelaer:
Perfect, thanks a lot,

Olivier

PR #211, opened 2025-12-17
Add support for specifying which `gzip` binary to use

GabrielMajeri:
I've looked at a recent simulation we did using MG5, and it took about 1 hour to generate the events, then another hour **just to compress the `.hepmc` file generated by Pythia**! This is because `gzip` is slow when it comes to compressing a single, large file (like an 80 GB `.hepmc` file), and from what I've checked, it doesn't parallelize its work.

This PR adds support for specifying which `gzip` binary to use. [`pigz`](https://github.com/madler/pigz) is a multithreaded implementation of gzip which is much faster than the vanilla one. I've made a small benchmark for generating 10.000 `p p > t tbar` events, and the results I got are:

```
Vanilla gzip:
real    6m14.368s
user    9m34.663s
sys     0m22.435s

Multithreaded pigz (with this PR) on a 128 core machine:
real    3m58.925s
user    9m44.710s
sys     2m26.692s
```

It went down from a wall time of 6 minutes to 4 minutes, a 33% improvement.

Please let me know how I could refactor this code to better fit the existing framework. I wasn't sure how to add a new option to MadGraph, so I just tried to follow the example for the `eps_viewer` setting.

One other thing we could do is check the `nb_core` option and pass the `-p` option to `pigz`, to limit the number of cores it uses. In a cluster setting, I'm not sure whether we could use multiple cores at all at this step of the job.

## Summary by Sourcery

Allow configuration of the gzip compression tool and integrate it into the run and MadGraph interfaces.

New Features:
- Add a configurable gzip tool option (e.g. pigz) to control which binary is used for file compression.

Enhancements:
- Lower the file size threshold for switching to external compression to improve performance on moderately large files.
- Wire gzip configuration into both run and MadGraph configuration flows so the chosen tool is consistently applied.

### DIFF
--- a/input/.mg5_configuration_default.txt
+++ b/input/.mg5_configuration_default.txt
@@ -32,7 +32,7 @@
 # acknowledged_v3.1_syntax = False
 
 
-#! Prefered Fortran Compiler
+#! Preferred Fortran Compiler
 #! If None: try to find g77 or gfortran on the system
 #!
 # fortran_compiler = None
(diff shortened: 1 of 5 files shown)


oliviermattelaer:
Hi,

Thanks for the PR, this is a really interesting one.
Now I would say in the configuration file that we support only gzip and pizg.
As important feature to add is indeed that we follow the nb_core constraint.
maybe one simple solution is to setup  "_gzip_tool" 
as a template command like either
"gzip -f %(path)s"
or 
"pizg -p 16 -f %(path)"

such that all the option to handle are handle in your helper routine 

Thanks,

Olivier

GabrielMajeri:
Thanks for the feedback. I've updated the PR accordingly.

I've replaced the `gzip` setting with an `use_pigz` setting (with three possible values: `None` - autodetect, `False` - stick to `gzip`, `True` - use `pigz`). I've also made it so that we respect the `nb_cores` option.

oliviermattelaer:
Sounds good to me.
One question before doing some manual test, would it make sense to use pigz for unzipping file as well?

Olivier

GabrielMajeri:
I don't know if it would help that much. The `pigz` [manual page](https://zlib.net/pigz/pigz.pdf) states:

> Decompression can’t be parallelized, at least not without specially prepared deflate streams for
> that purpose. As a result, pigz uses a single thread (the main thread) for decompression, but will
> create three other threads for reading, writing, and check calculation, which can speed up decom-
> pression under some circumstances. Parallel decompression can be turned off by specifying one
> process ( -dp 1 or -tp 1 ).

If you tell me where in the code I should look for `gunzip`, I will try to see if we can replace it with `pigz` and benchmark to see if it works any faster.

Furthermore, should I also update some documentation or change log somewhere, to let end users know how to enable this feature?

oliviermattelaer:
So let's not do it since the gain is likely too small and that we use some option of gunzip which might be broken by pigz.
For the documentation, I will add it to the help_set function, don't worry about it.
I will now do some run to validate in both mode and then add a CI test testing pigz, before merging this.

Thanks,

Olivier

oliviermattelaer:
Hi,

This is now merged in the 3.6.7 branch, which should be  merged soon in 3.x branch (and therefore released).

Thanks a lot,

Olivier

GabrielMajeri:
Great! Is it alright to close this PR/branch, then?

PR #212, opened 2025-12-18
Fix import error in tarball

cverstege:
Fix an import error when the madgraph module is named internal, e.g. in a tarball.

See https://answers.launchpad.net/mg5amcnlo/+question/823283

## Summary by Sourcery

Bug Fixes:
- Fall back to importing misc utilities from an internal module if the standard madgraph package import fails during IOError handling in file copy operations.


oliviermattelaer:
Ah our message cross .... 
So this is already in the pipeline for the LTS 2.9.27 and therefore will be included in the latest dev version (that you did target) in 3.6.7.

So closing this.

Thanks,

Olivier

PR #218, opened 2026-02-24
Patch for CPP writer to mach fortran for FDgauge

stloufra:
Just added 2 lines to match shifting of indices for scalar variables for FD gauge in CPP writer.

## Summary by Sourcery

Bug Fixes:
- Correct scalar variable index offsets in the C++ writer when using FD gauge with unitary_gauge == 3 to match the Fortran implementation.


oliviermattelaer:
Excellent, the Japanese are quite exited by your work!

PR #219, opened 2026-02-25
HotFix: Import logging before using it in ./bin/mg5_aMC

RHofsaess:
In case of python < 3.10, logging is used in the deprecation warning before being imported.
Easy fix is to move the import up.

@roiser

## Summary by Sourcery

Bug Fixes:
- Prevent failures in Python versions prior to 3.10 where logging was used in a deprecation warning before being imported.

### DIFF
--- a/bin/mg5_aMC
+++ b/bin/mg5_aMC
@@ -23,7 +23,6 @@ and call immediately the command line interface scripts"""
 import sys
 import logging
 import logging.config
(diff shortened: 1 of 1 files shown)


roiser:
thanks lot @RHofsaess , just to add that Robin was interested to have this in 3.6.7 for his GPU benchmarking work

oliviermattelaer:
Thanks a lot,

Olivier

Issue #223, opened 2026-03-26
Madgraph-Matchbox interface bug: calling the wrong Fortran template with split coupling orders


k-kak:
Hello,

I'm a Herwig developer working on adding BSM support to Matchbox via interfacing it with Madgraph as an external amplitude provider. Adding general UFO model support, our user might want to specify the coupling order at the squared amplitude level (NP^2 option). This will activate a different "mode of operation" of Madgraph than the normal one. The following is a summary of the problem along with a suggested fix to a bug that looks more of a typo really.

When Matchbox uses Madgraph as an external amplitude provider, the command "output matchbox -f" is run by Madgraph as part of matchbox's steering of Matchbox. This tells Madgraph to use its matchbox-specific Fortran output mode. In the normal case, Madgraph uses the template `matrix_standalone_matchbox.inc` to generate a `matrix.f `file per subprocess that defines Fortran subroutines that matchbox calls, whereas in the special case of split coupling orders, i.e. `len(split_orders)>0` in `madgraph/iolibs/export_v4.py`, I think Madgraph uses the wrong template file, `matrix_standalone_splitOrders_v4.inc` for LO and the correct template for NLO, `matrix_standalone_matchbox_splitOrders_v4.inc`.  I think both LO and NLO Matchbox interfaces with Madgraph should use the latter template. I also believe that this might be easily fixed by modifying the relevant `elif` condition to `elif self.opt[’export_format’] in [’ madloop_matchbox’,'matchbox']`. I'd be gratefull if this issue is resolved such that Madgraph can be integrated in the new Matchbox workflow.

Bests
Karim

oliviermattelaer:
Sure.
Is this what you had in mind?
https://github.com/mg5amcnlo/mg5amcnlo/commit/b15ab632e8ec6fa4f11adc4c088202e1d66eee91
Did you /can you test it?

Cheers,

Olivier

k-kak:
> Sure.
> Is this what you had in mind?
> https://github.com/mg5amcnlo/mg5amcnlo/commit/b15ab632e8ec6fa4f11adc4c088202e1d66eee91
> Did you /can you test it?
> 
> Cheers,
> 
> Olivier

[...]

PR #233, opened 2026-04-13
Fix: implicit typing bug in `GET_JAMP`/`GET_LNJAMP` (matchbox split-orders template)

k-kak:
This is related to #223. 

### Problem

When calling `GET_JAMP` / `GET_LNJAMP` from C (via Herwig's Matchbox interface), the array bounds check crashed with:                                                     

```
Index '0' of dimension 2 of array 'jamp' below lower bound of 1
```
### Cause

In `matrix_standalone_matchbox_splitOrders_v4.inc`, the subroutines `GET_JAMP` and `GET_LNJAMP` take `SOINDEX` as an argument but never declare it. `GET_JAMP` also references `LNJAMP` in its `COMMON` block without declaring it, producing a silent size mismatch against `MATRIX` / `BORN`.

### Fix

In both `GET_JAMP` and `GET_LNJAMP`:
- Add `SOINDEX` to the `INTEGER` declaration.
- Declare `LNJAMP(NCOLOR,NAMPSO)` alongside `JAMP` so the `COMMON` block matches the other subroutines for backwards compatibility.

## Note

I opened this PR targeting 3.5.14 branch. I am not sure if this is the correct procedure, but I am happy to adjust the target if needed.

## Summary by Sourcery

Fix implicit typing and common-block declarations in matchbox split-orders standalone matrix interface to prevent array bound errors when accessing JAMP data from C.

Bug Fixes:
- Declare SOINDEX as an integer argument in GET_JAMP and GET_LNJAMP to avoid out-of-bounds array access in JAMP when called from C.
- Declare LNJAMP in the GET_JAMP common block with the correct dimensions to match other subroutines and prevent size mismatches.

Enhancements:
- Normalize the name of the get_chosen_so_config helper subroutine to use the consistent proc_prefix-based naming convention.

### DIFF
--- a/madgraph/iolibs/template_files/matrix_standalone_matchbox_splitOrders_v4.inc
+++ b/madgraph/iolibs/template_files/matrix_standalone_matchbox_splitOrders_v4.inc
@@ -351,11 +351,11 @@ C ----------
 
       SUBROUTINE %(proc_prefix)sGET_JAMP(njamp, SOINDEX, ONEJAMP)
 
-      INTEGER     NCOLOR, NJAMP
+      INTEGER     NCOLOR, NJAMP, SOINDEX
       PARAMETER (NCOLOR=%(ncolor)d) 
       INTEGER NAMPSO
       PARAMETER (NAMPSO=%(nAmpSplitOrders)d)
-      COMPLEX*16  JAMP(NCOLOR,NAMPSO), ONEJAMP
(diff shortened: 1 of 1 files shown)


oliviermattelaer:
Thanks a lot ! will be in 3.5.15 (I can also already include it in 3.7.1 which will be release sooner)

PR #253, opened 2026-05-14
preserve negative iseed in run_card across runs

oliviermattelaer:
## Summary

- Allow a negative `iseed` in the `run_card` as an opt-out to the existing reset-to-0 behavior. A negative value is preserved on disk so subsequent runs reuse the same seed.
- The Fortran code still receives a non-negative seed: the absolute value is exported both to the `run_card.inc` include file and to the `randinit` file. This means `iseed = -5` produces the exact same effective seed as `iseed = 5` for a single run.
- Behavior is consistent at LO (madevent) and NLO (amc@nlo).

## Changes

- `madgraph/interface/common_run_interface.py` — `reset_iseed_in_run_card` only resets when `iseed > 0`, preserving negative values.
- `madgraph/interface/madevent_interface.py` — LO uses `abs(iseed)` for `self.random`.
- `madgraph/interface/amcatnlo_run_interface.py` — NLO writes `abs(iseed)` to `randinit` when negative.
- `madgraph/various/banner.py` — adds `mod_inc_iseed` on `RunCard` so the include file always gets `abs(iseed)` (covers both LO and NLO via the existing `mod_inc_<key>` mechanism).
- `tests/unit_tests/various/test_banner.py` — new `test_negative_iseed` covers both LO and NLO: include-file export, on-disk preservation of negative iseed, and unchanged reset for positive iseed.
- `.github/workflows/unittest.yml` — wires the new test into `unittest_15`.

## Test plan

- [x] `./tests/test_manager.py test_negative_iseed -t0` passes.
- [x] All 29 existing tests in `test_banner.py` still pass.
- [x] CI (`unittest_15`) green on this branch.

## Summary by Sourcery

Allow negative random seeds in run cards to be preserved across runs while still providing non-negative seeds to the Fortran backends, with consistent behavior at LO and NLO.

Bug Fixes:
- Prevent unintended resetting of user-chosen negative random seeds in run cards by only zeroing positive seeds between runs.

Enhancements:
- Export the absolute value of the random seed from run cards to the Fortran include files and NLO randinit file so negative seeds are handled transparently.
- Unify random seed handling between LO and NLO interfaces so they both interpret negative seeds as reusable while passing their absolute value to the generator.

CI:
- Include the new negative-seed behavior test in the unittest_15 GitHub Actions workflow.

Tests:
- Add a unit test covering negative iseed behavior for both LO and NLO run cards, including include-file export and on-disk persistence.

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -326,10 +326,10 @@ jobs:
       - uses: actions/checkout@v5
 
       # Runs a set of commands using the runners shell
-      - name: test one of the test test_autodef_nomissmatch test_basic test_check_valid_LO test_custom_fcts test_default test_fixed_fac_scale_block test_guess_entry_fromname test_pdlabel_block test_analyse_card_analyse test_analyse_card_default test_create_param_dict test_define_not_dep_param test_order_param test_write_block test_write_param test_write_qnumber test_define_not_dep_param test_full_write test_write_orders test_write_particles test_write_vertices test_add_external_parameters test_add_particle test_couplings test_identify_particle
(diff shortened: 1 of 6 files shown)


marcozaro:
Hi,
i tried what follows.
- 1st run, set iseed =10
- 2nd run, send another run without changing the run card
- 3rd run set iseed =-10
- 4th run, send another run without changing the run card
the input is

generate p p > e+ ve [QCD]
output
launch NLO
set iseed 10

launch NLO

launch NLO
set iseed -10

launch NLO

I get 1=3=4, 2!=1, as expected

Marco

marcozaro:
so it can be merged, thanks a lot

PR #256, opened 2026-05-15
Drop cached model package in load_model so internals re-import

oliviermattelaer:
## Summary
- `models.load_model` wipes `sys.modules['object_library']` and other internal modules on every call, then re-imports the model package. When that package is already cached (e.g. a second `generate` of the same model), the `__import__` is a no-op, `__init__.py` never re-runs, and the internal modules stay deleted.
- Live `Function`/`Lorentz`/`Particle` instances then hold a `__class__` whose `__module__` is no longer reachable through `sys.modules`. Pickle can't represent the class, which breaks the multiprocessing pool in `FKSHelasMultiProcess` under `low_mem_multicore_nlo_generation` — surfacing as `_pickle.PicklingError: Can't pickle <class 'object_library.Function'>: import of module 'object_library' failed`.
- Fix: also drop the cached model package itself before the re-import so `__init__.py` re-executes and consistently repopulates the internals.

## Test plan
- [x] `./tests/test_manager.py -t0 test_wj_loonly_gen` — was failing, now passes
- [x] `./tests/test_manager.py -t0 test_z_nlo_gen_qcd` — still passes
- [ ] Smoke-test other NLO flows that load the same model twice in one session

## Summary by Sourcery

Bug Fixes:
- Avoid leaving model internals orphaned in sys.modules when reloading a model so live objects remain serializable across multiprocessing.

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -393,10 +393,10 @@ jobs:
       - uses: actions/checkout@v5
 
       # Runs a set of commands using the runners shell
-      - name: test one of the test testIO_test_pptt_fksreal testIO_test_pptt_fksrealew testIO_test_tdecay_fksreal testIO_test_wprod_fksew test_w_nlo_gen_gosam test_w_nlo_gen_qcd test_w_nlo_gen_qed test_wj_loonly_gen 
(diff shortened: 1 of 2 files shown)


oliviermattelaer:
This creates more side effect than fixing -> drop for now and we will likely revisit such type of change when the branch is more stable

PR #293, opened 2026-05-27
Also look in new env variable MG5_LIBRARY_PATH for libraries.

APN-Pucky:
This avoids polluting commonly used global variables. Further the error message for missing CTlib and CTmod is split for more information in case of a failure.

This makes it much easier to distribute a working pre-compiled madgraph on linux distributions with `export MG5_LIBRARY_PATH="/usr/lib64/:/usr/include/mpfun90"` without having to set a commonly used global env variable.

For instance I use:
```
my_process/lib $ ls -lah libcts.a libiregi.a mpmodule.mod
lrwxrwxrwx 1 apn apn 19 May 27 16:05 libcts.a -> /usr/lib64/libcts.a
lrwxrwxrwx 1 apn apn 21 May 27 16:05 libiregi.a -> /usr/lib64/libiregi.a
lrwxrwxrwx 1 apn apn 33 May 27 16:05 mpmodule.mod -> /usr/include/mpfun90/mpmodule.mod
```

## Summary by Sourcery

Add support for locating required CutTools libraries via a dedicated MG5_LIBRARY_PATH environment variable and improve error reporting when required files are missing.

Enhancements:
- Use the MG5_LIBRARY_PATH environment variable in library search paths to avoid relying only on generic system-wide variables.
- Provide more specific error messages when either libcts.a or mpmodule.mod cannot be found in the environment paths.


oliviermattelaer:
Thanks this is a cool addition,

Cheers,

Olivier

oliviermattelaer:
CI on 3.7.2 confirmed that this is not problematic

PR #324, opened 2026-07-22
change MadSpin defaults: spinmode=madspin, density_keep_jacobian=True…

oliviermattelaer:
…, sequential_decay=auto

- spinmode default goes back to 'madspin' (now the density off-shell implementation) instead of 'PA'.
- density_keep_jacobian defaults to True: the PA reshuffling jacobian is folded into the accept/reject weight instead of being applied as a post-acceptance kinematic dressing.
- sequential_decay defaults to 'auto' (via the ConfigFile auto mechanism): resolved at run time to sequential accept/reject for the PA/onshell pole approximations and to the joint test for madspin/full. fixed_order and unsupported spinmodes still force the joint test.
- update the default madspin card and stale comments, add unit tests for the auto gate and the new defaults.

## Summary by Sourcery

Update MadSpin defaults and configuration handling for spin mode, jacobian treatment, and sequential decay behaviour.

New Features:
- Introduce an automatic sequential_decay mode that selects sequential or joint accept/reject based on the chosen spinmode.

Enhancements:
- Change the default spinmode to madspin to use the off-shell matrix element plus density configuration.
- Enable density_keep_jacobian by default so the reshuffling jacobian is included in the accept/reject weight instead of applied as a post-acceptance kinematic dressing.
- Make sequential_decay default to the auto configuration via the ConfigFile mechanism, with fixed_order continuing to enforce the joint test.
- Refresh the default MadSpin card comments to reflect the new spinmode and jacobian defaults.

Tests:
- Add unit coverage for the sequential_decay auto gate across spinmodes and fixed_order, and for the new MadSpin option defaults.
- Adjust parallel test expectations and comments to align with the updated default spinmode behaviour.

### DIFF
--- a/MadSpin/interface_madspin.py
+++ b/MadSpin/interface_madspin.py
@@ -64,7 +64,7 @@ class MadSpinOptions(banner.ConfigFile):
         self.add_param('ms_dir', '')
         self.add_param('max_running_process', 100)
         self.add_param('onlyhelicity', False)
-        self.add_param('spinmode', "PA", allowed=['full', 'madspin', 'none', 'onshell', 'PA', 'madspin_v1', 'onshell_v1'])
+        self.add_param('spinmode', "madspin", allowed=['full', 'madspin', 'none', 'onshell', 'PA', 'madspin_v1', 'onshell_v1'])
         self.add_param('use_old_dir', False, comment='should be use only for faster debugging')
         self.add_param('run_card', '' , comment='define cut for decay_events (in onshell frame). Path to run_card to use')
         self.add_param('fixed_order', False, comment='to activate fixed order handling of counter-event')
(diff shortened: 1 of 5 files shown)


MadSpin/interface_madspin.py
```diff
@@ -80,8 +80,11 @@ def default_setup(self):
...
-        self.add_param('sequential_decay', False, comment='accept/reject one decaying particle at a time instead of the full set at once (density mode). Exact and much cheaper when several particles decay; set to False for the historical joint accept/reject.')
+        self.add_param('density_keep_jacobian', True, comment='PA spinmode only: fold the offshell-reshuffling phase-space jacobian into the accept/reject weight (default) instead of applying the reshuffle as a post-acceptance kinematic dressing (False). Ignored by the madspin/full spinmodes, which always include that jacobian.')
+        self.add_param('sequential_decay', False, comment='accept/reject one decaying particle at a time instead of the full set at once (density mode). Exact and much cheaper when several particles decay. Default is auto: True for the PA/onshell spinmodes, False (joint accept/reject) for madspin/full.')
```
sargyro, line 84: @oliviermattelaer is this always faster? If so why don't we turn it on by default everywhere?

PR #325, opened 2026-07-23
CI: run the unit and acceptance tests that no workflow triggered

oliviermattelaer:
## Summary

Port to `3.7.3` of the CI coverage audit validated on 3.x (branch `claude/untriggered-test-functions-cf1868`, all workflows green there). Every CI job lists its tests explicitly, and an audit of the workflows against the test suite showed that a number of unit and acceptance tests were listed nowhere, so they never ran.

### unittest.yml
- new `unittest_uncovered` job with 19 previously untriggered unit tests (color algebra, EW sudakov generation + IOTests, `testIO_test_ppw_fksall`, `testIO_Loop_sqso_uux_ddx`, import_ufo symmetric color/lorentz, restrict model, process checks, shower card, ...)
- new `unittest_write_model` job: `test_write_model` runs isolated — it only passes when no other test has loaded a UFO model first (the known "side effect" that had it excluded)
- `test_fks_ppzz_in_RS` stays out: the RS UFO model is still python2-only

### acceptancetest.yml
- re-enabled the commented-out jobs for `test_gen_evt_onlygen`, `test_generate_events_lo_hw6_stdhep`, `test_generate_events_lo_py6_stdhep` and `test_madspin_LOonly` (now with `restore_all`)
- new jobs for the four `check gauge`/`check full` tests, `test_decay_chain_identical_particle_outoforder` and `test_madevent_dy3j_mlm`
- still excluded, documented in place: `test_contur_from_file` (needs a configured `contur_path` on the runner), the two eva tests (assert run_card defaults that do not match the current ones; `test_ieva_collision` also uses an undefined variable), and the madweight tests (excluded on request)

### Fixes required for the above to pass (all validated on the 3.x run)
- **export_v4**: normalise negative zeros when writing JAMP coefficients — python 3.14's C99 complex semantics make the sign of a zero version-dependent — and update the `test_pptt_fksreal` reference accordingly
- `test_write_model`: add the model directory itself to `sys.path` (UFO models use implicit relative imports)
- `test_get_symmetric_lorentz`: don't require the exact SSVV index; the process-global UFO module can already contain an equivalent lorentz which is legitimately reused
- stdhep tests: `generate_events aMC@LO` names the run `run_01`; the `_LO` suffix only applies to fixed-order LO mode

Note: the PA-decay/wj-madspin tests and the density jobs exist only on 3.x (madspin work) and are not part of this port; they reach 3.7.x whenever 3.x is merged.

## Test plan
- [x] the 19-test `unittest_uncovered` batch, `test_write_model` and `testIO_test_pptt_fksreal` pass locally on this branch (3.7.3 base, python 3.14)
- [x] the same content is fully green on 3.x (unittest, acceptance, aloha, parallel workflows)
- [ ] CI on this PR (acceptance jobs need warm HEPTools caches — if many jobs fail within seconds, check for "Cache not found" and re-run after warming)

## Summary by Sourcery

Expand CI coverage to run previously untriggered unit and acceptance tests and make small fixes so they pass across supported Python versions.

Enhancements:
- Add unittest jobs to run uncovered tests and isolate the side-effectful test_write_model case.
- Re-enable and add acceptance test jobs for stdhep event generation, madspin, gauge/full checks, decay-chain ordering and DY+3j MLM matching, with appropriate cache restores and conditions.
- Document and keep disabled acceptance tests that require unavailable external tools or rely on outdated run_card defaults.
- Normalize complex-number formatting in export_v4 to avoid Python-version-dependent negative zeros and update the corresponding reference matrix file.
- Relax UFO symmetric Lorentz structure assertions and adjust write_model imports to work with shared global UFO modules and implicit relative imports.
- Align stdhep acceptance tests with the actual run directory naming used by aMC@LO event generation.

### DIFF
--- a/.github/workflows/acceptancetest.yml
+++ b/.github/workflows/acceptancetest.yml
@@ -1040,8 +1040,9 @@ jobs:
 
       - name: run the test
         run: |
-            #./tests/test_manager.py test_contur_from_file test_rivet_from_file -pA -t0 -l INFO 
-            PYTHONPATH=$PYTHONPATH:/home/runner/.cache/HEPtools/contur/python3.10:/home/runner/.cache/HEPTools/rivet/local/lib/python3.10/dist-packages/:/home/runner/.cache/HEPTools/yoda/local/lib/python3.10/dist-packages/ LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/home/runner/.cache/HEPTools/lib ./tests/test_manager.py  test_rivet_from_file -pA -t0 -l INFO  
+            # test_contur_from_file disabled: needs the mg5 contur_path option pointing to a full contur install (conturenv.sh), not available on the runner
+            #./tests/test_manager.py test_contur_from_file test_rivet_from_file -pA -t0 -l INFO
+            PYTHONPATH=$PYTHONPATH:/home/runner/.cache/HEPtools/contur/python3.10:/home/runner/.cache/HEPTools/rivet/local/lib/python3.10/dist-packages/:/home/runner/.cache/HEPTools/yoda/local/lib/python3.10/dist-packages/ LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/home/runner/.cache/HEPTools/lib ./tests/test_manager.py  test_rivet_from_file -pA -t0 -l INFO
             
   acceptancetest_contur2:
     # The type of runner that the job will run on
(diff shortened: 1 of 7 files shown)


oliviermattelaer:
Wrong target: this work is based on madspin_density, superseded by the PR into madspin_density. The 3.7.3 port branch (untriggered-tests-3.7.3) stays available if a direct 3.7.3 backport is wanted later.

PR #332, opened 2026-08-07
Quarkonium and leptonium production

lsimon125:
Quarkonium and leptonium production including S- and P-wave states, based on [arXiv:2510.26773](https://arxiv.org/pdf/2510.26773) and [arXiv:2607.26739](https://arxiv.org/pdf/2607.26739).

- Introduced a dedicated UFO model format providing the necessary Fock states. The corresponding SM implementation, `sm_onia`, is available in the database.

- Introduced dual numbers, provided as an independent FORTRAN module, for automatic differentiation.

## Summary by Sourcery

Add infrastructure and physics support for quarkonium and leptonium (non-relativistic bound states) production with S- and P-wave configurations, including automatic differentiation using dual numbers and corresponding onia-specific cards, reshuffling, projectors and tests.

New Features:
- Support quarkonium and leptonium bound states via an onia-aware UFO model and corresponding MadGraph export paths.
- Introduce dual-number based HELAS library variants and differentiation utilities for P-wave onia amplitudes.
- Add onia-specific configuration (onia_card, LDME handling, momentum reshuffling) into run-time cards and interfaces.

Enhancements:
- Extend phase-space generation, color and helicity handling, and matrix-element construction to correctly treat non-relativistic bound states.
- Wire Ninja loop reduction to use quadruple precision where available and adjust loop interface configuration.
- Improve multiparticle defaults and polarization handling to account for onia models, and refine logging and documentation of options.

Documentation:
- Document new momentum reshuffling options and onia-specific run-card blocks for MadGraph users.

Tests:
- Add acceptance and standalone tests for quarkonium and leptonium production to validate new onia and dual-number infrastructure.
- Update existing loop-induced and MatchBox IO comparison tests to reflect the new Ninja quadruple-precision behavior.


Template/LO/SubProcesses/setcuts.f
```diff
@@ -12,6 +12,7 @@ SUBROUTINE SETCUTS
       include 'coupl.inc' ! needs VECSIZE_MEMMAX (defined in vector.inc)
       include 'run.inc'
       include 'cuts.inc'
+      include 'ldme.inc'
 c
```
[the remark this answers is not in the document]
oliviermattelaer: fixed on the export side.

madgraph/core/diagram_generation.py
```diff
@@ -1738,11 +1746,11 @@ def generate_multi_amplitudes(cls,process_definition,
                 if not all(istags):
                     raise MadGraph5Error("Tagging only one initial-state particle is not allowed")
                 islegs = [\
-                        fks_tag.TagLeg({'id':id, 'state': False, 'polarization': isleg['polarization'], 'is_tagged': tag}) \
+                        fks_tag.TagLeg({'id':id, 'state': False, 'polarization': isleg['polarization'], 'onium': islegs['onium'], 'is_tagged': tag}) \
                         for id, isleg, tag in zip(prod, islegs_orig, istags)]
             else:
```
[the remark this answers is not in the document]
oliviermattelaer: fixed in b1d344d: `'onium': islegs['onium']` is now `'onium': isleg['onium']`,

madgraph/core/diagram_generation.py
```diff
@@ -159,6 +159,11 @@ def vertices_from_link(cls, link, model, first_vertex = False):
         legs = base_objects.LegList(sorted([l for l,v in leg_vertices],
                                            key= lambda l: l.get('number'), reverse=True))
 
+        # LS::ONIUM PROPERTIES ARE MISSING IN link.links
+        # for i in range(len(legs)):
+        #     if (legs[i].get('number')==3) or (legs[i].get('number')==4):
+        #         legs[i].set('onium',{'name': 'my_hardcoded_onium'})
+
```
oliviermattelaer: We should clean this. 
Like this it is reading that something is currently missing 
(and we should avoid commented code) but for good commented reason

madgraph/core/diagram_generation.py
```diff
@@ -236,11 +241,14 @@ def vertex_from_link(cls, legs, vertex_id, model):
     def leg_from_link(link):
         """Return a leg from a link"""
 
+        # LS:: ONIUM PROPERTIES HAVE TO BE CHECKED
```
oliviermattelaer: same here. clean such comment

madgraph/core/diagram_generation.py
```diff
@@ -236,11 +241,14 @@ def vertex_from_link(cls, legs, vertex_id, model):
...
         if link.end_link:
             # This is an external leg, info in links
             return base_objects.Leg({'number':link.links[0][1],
                                      'id':link.links[0][0][0],
                                      'state':(link.links[0][0][1] == 0),
+                                     # 'onium':link.links[0][0][2],
+                                     'onium': {},
```
oliviermattelaer, line 244: This also is not clean

madgraph/core/helas_objects.py
```diff
@@ -505,10 +505,17 @@ def leg_from_link(link):
         """Return a leg from a link"""
 
         if link.end_link:
+            # LS::ONIUM PROPERTIES ARE MISSING IN link.links
+            # if (link.links[0][1][0]==3) or (link.links[0][1][0]==4):
+            #     onium = {'name':'my_onium'}
+            # else:
+            #     onium = {}
```
oliviermattelaer: same here

madgraph/core/helas_objects.py
```diff
@@ -4814,14 +4841,84 @@ def get_number_of_amplitudes(self):
...
+             except: 
+                pass
+        if npwave%2:
+            raise MadGraph5Error("Odd number of quarkonia constituents detected.")
+        else:
+            npwave /= 2
+
+        return int(npwave)
+    
+    # LM::This function should evaluate what is the higher derivative order needed. For the time being I have introduced a placeholder function
```
oliviermattelaer: comment does not seem up-to-date anymore (is it?)

madgraph/core/helas_objects.py
```diff
@@ -4814,14 +4841,84 @@ def get_number_of_amplitudes(self):
         return sum([ len(d.get('amplitudes')) for d in \
                        self.get('diagrams')])
 
+    def get_nonia(self):
```
oliviermattelaer, line 4844: Is this function call many times?
This might be a quite expansive function to run (you loop over all wavefunctions)
so it might be worthed to cache the result (rather than recomputing it)
(but just a suggestion)

madgraph/fks/fks_common.py
```diff
@@ -43,11 +43,11 @@ class FKSDiagramTag(diagram_generation.DiagramTag): #test written
 
     @staticmethod
     def link_from_leg(leg, model):
-        """Returns the default end link for a leg: ((id, number), number).
+        """Returns the default end link for a leg: ((id, number, onium), number).
```
oliviermattelaer, line 46: Is this needed?

madgraph/interface/madgraph_interface.py
```diff
@@ -6149,14 +6293,30 @@ def add_default_multiparticles(self):
...
             b = self._curr_model.get_particle(5)
+            c = self._curr_model.get_particle(4)
             if not b:
                 break
+            if not c:
+                break
 
+            if 4 in multi:
+                if c['mass'] != 'ZERO':
+                    if 'onia' in self._curr_model.get('name'):
```
oliviermattelaer, line 6326: Would it make more sense to always do it? (and do a consistent setup for 3 flavor)
and not only in the onia mode?

madgraph/interface/madgraph_interface.py
```diff
@@ -9828,40 +10018,50 @@ def finalize(self, nojpeg, online = False, flaglist=[]):
...
+                        aloha.dual_mode = True
+                    self._curr_exporter.convert_model(self._curr_model,
+                                                   wanted_lorentz,
+                                                   wanted_couplings,
+                                                   npwave)
                 if hasattr(self, '_me_curr_exporter') and self._me_curr_exporter:
-                    self._me_curr_exporter.convert_model(self._curr_model, 
+                    self._me_curr_exporter.convert_model(self._curr_model,
                                                wanted_lorentz,
                                                wanted_couplings)
```
oliviermattelaer, line 10069: This sounds dangerous since looks like some exporter have npwave and not others. Would it make sense to have self._curr_exporter.npawe ?
(this would avoid such issue) 

or if possible use keyword arguments (such that one just need to enfore that all exporter use **kwargs at then end and ignore argument not for them

madgraph/iolibs/export_v4.py
```diff
@@ -1252,7 +1616,7 @@ def get_color_data_lines(self, matrix_element, n=128):
         rows in chunks of size n."""
 
         if not matrix_element.get('color_matrix'):
-            return ["DATA Denom/1/", "DATA CF/1/"]
+            return ["DATA Denom/1/", "DATA CF/0/"]
```
oliviermattelaer: This does worry me... 
Why do you change such value?

madgraph/iolibs/export_v4.py
```diff
@@ -1706,17 +2079,17 @@ def format(frac):
...
             
             if frac not in  [1., -1]:
-                res_list.append(' TMP_JAMP(%d) = %s + (%s) * %s ! used %d times' % (i,amp1, format(frac), amp2, nb))                
+                res_list.append((' TMP_JAMP(%d) = %s + (%s) * %s ! used %d times' % (i,amp1, format(frac), amp2, nb)).replace('_percent_', '%'))
             elif frac == 1.:
-                res_list.append(' TMP_JAMP(%d) = %s +  %s ! used %d times' % (i,amp1, amp2, nb))  
+                res_list.append((' TMP_JAMP(%d) = %s +  %s ! used %d times' % (i,amp1, amp2, nb)).replace('_percent_', '%'))
             else:
-                res_list.append(' TMP_JAMP(%d) = %s - %s ! used %d times' % (i,amp1, amp2, nb))  
+                res_list.append((' TMP_JAMP(%d) = %s - %s ! used %d times' % (i,amp1, amp2, nb)).replace('_percent_', '%'))
```
oliviermattelaer: This indicates issue with the previous formatting, you should use %% to avoid that (I guess)

madgraph/iolibs/export_v4.py
```diff
@@ -7206,6 +8150,9 @@ def copy_standard_file(self):
             load_card = ''
             lha_read_filename='lha_read.f'
             updateloop_default = '.false.'
+
+        cp( MG5DIR + '/models/template_files/fortran/rw_onia.f', \
+                                       os.path.join(self.dir_path,'rw_onia.f'))
```
oliviermattelaer, line 8279: Do we need to  always have that file (and to compile it).
Can we gate it to have it only when needed?

madgraph/iolibs/export_v4.py
```diff
@@ -1252,7 +1616,7 @@ def get_color_data_lines(self, matrix_element, n=128):
         rows in chunks of size n."""
 
         if not matrix_element.get('color_matrix'):
```
oliviermattelaer, line 1627:
```suggestion
if not matrix_element.get('color_matrix'):
    if matrix_element.get_nonia() > 0:
        return ["DATA Denom/1/", "DATA CF/0/"]   # colour-forbidden Fock state
    return ["DATA Denom/1/", "DATA CF/1/"]       # unchanged for everything else
```

madgraph/various/banner.py
```diff
@@ -4811,7 +4831,7 @@ def create_default_for_process(self, proc_characteristic, history, proc_def):
                     self.display_block.append('pdlabel')
 
             if any(i in beam_id for i in [1,-1,2,-2,3,-3,4,-4,5,-5,21,22]):
-                maxjetflavor = max([4]+[abs(i) for i in beam_id if  -7< i < 7])
+                maxjetflavor = max([3]+[abs(i) for i in beam_id if  -7< i < 7])
```
oliviermattelaer, line 4970: need to be consistent with the multiparticle default behavior

models/template_files/fortran/makefile_madevent
```diff
@@ -48,6 +48,11 @@ clean:
 
 couplings.o: ../maxparticles.inc ../run.inc ../vector.inc
 couplings2.o: ../vector.inc
+rw_onia.o: onia
+
+onia:
+	touch ldme.inc
+	touch onia_read.inc
```
oliviermattelaer, line 57: why? should not be:
rw_onia.o: ldme.inc onia_read.inc

this rules sounds that you recompile too often

models/template_files/fortran/makefile_standalone
```diff
@@ -42,6 +42,11 @@ clean:
 	$(RM) *.o $(LIBDIR)$(LIBRARY)
 
 couplings.o: ../maxparticles.inc ../run.inc ../cuts.inc
+rw_onia.o: onia
+
+onia:
+	touch ldme.inc
+	touch onia_read.inc
```
oliviermattelaer, line 51: same

Template/LO/Cards/run_card.dat
```diff
@@ -61,7 +61,7 @@ $fixed_fact_scale
  %(dsqrt_q2fact2)s  = dsqrt_q2fact2    ! fixed fact scale for pdf2
  %(dynamical_scale_choice)s = dynamical_scale_choice ! Choose one of the preselected dynamical choices
  %(scalefact)s  = scalefact        ! scale factor for event-by-event scales
-$RUNNING
+$RUNNING$mom_resh
```
oliviermattelaer: This needs to be set on his own lines, otherwise this will create side effects.

Template/LO/SubProcesses/unwgt.f
```diff
@@ -735,6 +736,12 @@ SUBROUTINE write_leshouche(p,wgt,numproc,do_write_events, ihel, icol, ivec)
 c     Add info on resonant mothers
 c     recall onbw since that might have configured onBW for the wrong config (check tt~a ,t >... for checking impact 
       call cut_bw(p)
+c     check if process contains onia (the file is only created in this case)
+c     if the process contains onia set icol to a negative value to force
+c     the flag 'is_LC' set to be false in the routine addmothers to bypass 
+c     the writing of intermediate particles
+      inquire( file=trim('../onia.inc'), exist=exists )
+      if (exists) icol = -abs(icol)
```
oliviermattelaer: This is a bad mechanism since it will creates a lot of IO (one check per event) better use a common block associated to onia than a IO check (in particular when run so many times)

oliviermattelaer (changes requested):
Here is my review (nothing to deep, mainly some cleaning) 
I can likely do most myself, but on one or two it might be better that you do it to ensure a proper handling (in particular the change to the unweight.f which is quite critical)

- I found no guard refusing onia at NLO. There is raise self.InvalidCmd("initial particles cannot be onia"), but nothing checks for an onium in a [QCD] process.

- common_run_interface.py:1183 CommonRunCmd.detect_card_type has no onia entry in to_search and no onia_card.dat return branch. The PR's diff on that file touches only the AskforEditCard machinery (all_card_name, to_init_card, init_onia, do_set, write_onia_card) — nothing in detect_card_type.

- reload_card has no onia branch either — common_run_interface.py:8082. After the user edits onia_card.dat in an editor, it falls to else: logger.debug('not keep in sync'), so the in-memory self.onia_card is stale for any subsequent set. Worth fixing at the same time, since do_set re-reads it explicitly (line 6248) — so it's clearly meant to stay in sync.

- Declare npwave = [0] in aloha/__init__.py and drop the hasattr guards, then extend both set_global copies (madgraph/various/misc.py:2186
tests/IOTests.py:58)
with dual_mode and npwave (the latter needs a copy, not a reference, since it's a list). That's four files, no physics change.

- Also worth flagging while you're in there: the variable is typed three ways. Declared False, incremented as a counter (aloha.dual_mode += 1), assigned True/False elsewhere, and consumed as a wave count at :5486 (aloha.npwave.append(aloha.dual_mode)). A save/restore fixes the leak but not the ambiguity.

Template/LO/SubProcesses/unwgt.f
lsimon125: I use now the precompiler to carry out the check at compilation time

models/template_files/fortran/makefile_madevent
lsimon125: This rule just creates the two files as empty files in case no onium is present. It doesn't compile them

madgraph/iolibs/export_v4.py
lsimon125: Since this file is part of the Source folder it shouldn't know anything about the specific process. If we only want to compile it when needed, we would have to change the logic such that the Source folder compiles differently for quarkonium and non-quarkonium processes

madgraph/iolibs/export_v4.py
LMaxia: The problem is that AMP_FORMAT is used in different parts of the code, and there is no common number of percent signs that works. We have then used this workaround to ensure that the formatting does not interact with % at all during the build phase, and we replace it with the actual sign only at the end.
It is not nice to see, I agree, but I think the one proposed is the simplest solution.

madgraph/core/helas_objects.py
lsimon125: This function is called a few times, but not that often that it leads to measurable slow down. I kept the function without a caching system to treat it the same way as `get_nexternal()` which follows a similar logic.

madgraph/core/helas_objects.py
LMaxia: Indeed, I missed this comment during the cleanup before merging. I should have removed it from the file.

PS: I also realised that an old comment was present in the dual_variables.f module. I have also removed the line in the same commit (it does not affect the output of the code).

models/template_files/fortran/makefile_madevent
oliviermattelaer: Following up on the `rw_onia.o: onia` question -- the empty-file creation is right, but the rule does force a rebuild on every `make`: `onia` is a phony target that never exists as a file, so it is always considered out of date and `rw_onia.o` is recompiled each time.

Measured on a fresh `p p > e+ e-` output: with the current rule a second `make` in `Source/MODEL` recompiles `rw_onia.o`; with the rule below it prints `Nothing to be done for 'all'`.

Depending on the two include files keeps exactly the behaviour you wanted (they are still created empty when absent, via the same idiom already used for `../run.inc` just below) and gives real dependency tracking.

```suggestion
rw_onia.o: ldme.inc onia_read.inc

ldme.inc:
	touch ldme.inc

onia_read.inc:
	touch onia_read.inc
```

models/template_files/fortran/makefile_standalone
oliviermattelaer: Same as for `makefile_madevent`: `onia` is a phony target, so `rw_onia.o` is rebuilt on every `make`. Depending on the two include files instead keeps the "create them empty when absent" behaviour (same idiom as `../maxparticles.inc` / `../run.inc` below) and stops the needless recompilation.

```suggestion
rw_onia.o: ldme.inc onia_read.inc

ldme.inc:
	touch ldme.inc

onia_read.inc:
	touch onia_read.inc
```

madgraph/iolibs/export_v4.py
oliviermattelaer: > there is no common number of percent signs that works

That is true only because of one line: inside the JAMP loop `res` is fed back in **as part of the format string**

```python
res = (res + "%s" + AMP_format) % (self.coeff(...), str(amp_number))
```

so every term re-scans the text produced by the previous terms. A `%%` becomes a real `%` on the first pass and then blows up on the second (`unsupported format character 'C'`), which is exactly why no fixed number of `%` works and why the `_percent_` placeholder is needed.

If the already-substituted part is not re-scanned -- format only the new fragment, then concatenate -- the problem disappears and a single `%%` is correct at every call site. It is a one-character change on each of the two lines (`res = (res + ...) % (...)` -> `res = res + (...) % (...)`):

```python
                    if common_factor:
                        res = res + ("%s" + AMP_format) % \
                                                   (self.coeff(coefficient[0],
                                                   coefficient[1] / abs(coefficient[1]),
                                                   coefficient[2],
                                                   coefficient[3]),
                                                   str(amp_number))
                    else:
                        res = res + ("%s" + AMP_format) % (self.coeff(coefficient[0],
                                                   coefficient[1],
                                                   coefficient[2],
                                                   coefficient[3]),
                                                   str(amp_number))
```

(these two lines are outside the PR diff so GitHub will not let me attach them as a suggestion -- the three suggestions below cover the rest)

Checked that it round-trips over several terms:

```
AMP_format = "AMP(%s)" + "%%COMP(2**NPWAVE-1)"
-> JAMP(1)=+AMP(1)%COMP(2**NPWAVE-1)-AMP(2)%COMP(2**NPWAVE-1)+AMP(3)%COMP(2**NPWAVE-1)
```

`generate g g > hc(1|1P11) g` and the three onia acceptance tests still pass with `_percent_` fully removed.

**The four changes have to go in together** -- with `%%` in place but the accumulator still re-scanning, the build raises `ValueError`.

madgraph/iolibs/export_v4.py
```diff
@@ -1648,7 +2021,7 @@ def get_JAMP_lines(self, col_amps, JAMP_format="JAMP(%s)", AMP_format="AMP(%s)",
     
                 if common_factor:
                     res = res + ')'
-                res_list.append(res)
+                res_list.append(res.replace('_percent_', '%'))
```
oliviermattelaer: 1/3 of the `_percent_` removal (see the reasoning above -- needs the two accumulator lines changed as well).

```suggestion
                res_list.append(res)
```

madgraph/iolibs/export_v4.py
```diff
@@ -1706,17 +2079,17 @@ def format(frac):
...
             else:
-                res_list.append(' TMP_JAMP(%d) = %s - %s ! used %d times' % (i,amp1, amp2, nb))  
+                res_list.append((' TMP_JAMP(%d) = %s - %s ! used %d times' % (i,amp1, amp2, nb)).replace('_percent_', '%'))
 
         jamp_res = collections.defaultdict(list)
         max_jamp=0
         for (jamp, var), factor in new_mat.items():
             if var > 0:
-                name = AMP_format % var
+                name = (AMP_format % var).replace('_percent_', '%')
```
oliviermattelaer: 2/3 of the `_percent_` removal. Here `amp1`/`amp2`/`name` are *arguments*, so a real `%` inside them is never re-scanned -- once `AMP_format` carries `%%` these unescapes are all redundant.

```suggestion
                res_list.append(' TMP_JAMP(%d) = %s + (%s) * %s ! used %d times' % (i,amp1, format(frac), amp2, nb))
            elif frac == 1.:
                res_list.append(' TMP_JAMP(%d) = %s +  %s ! used %d times' % (i,amp1, amp2, nb))
            else:
                res_list.append(' TMP_JAMP(%d) = %s - %s ! used %d times' % (i,amp1, amp2, nb))

        jamp_res = collections.defaultdict(list)
        max_jamp=0
        for (jamp, var), factor in new_mat.items():
            if var > 0:
                name = AMP_format % var
```

madgraph/iolibs/export_v4.py
```diff
@@ -5052,9 +5788,15 @@ def write_matrix_element_v4(self, writer, matrix_element, fortran_model,
 
         # Extract JAMP lines
         # If no split_orders then artificiall add one entry called 'ALL_ORDERS'
-        jamp_lines, nb_temp = self.get_JAMP_lines_split_order(\
-                             matrix_element,amp_orders,split_order_names=
-                        split_orders if len(split_orders)>0 else ['ALL_ORDERS'])
+        if aloha.dual_mode:
+            jamp_lines, nb_temp = self.get_JAMP_lines_split_order(\
+                                matrix_element,amp_orders,
+                                AMP_format_addon= "_percent_COMP(2**NPWAVE-1)",
```
oliviermattelaer: 3/3 of the `_percent_` removal: with the accumulator no longer re-scanning, the escaped `%%` is enough here.

```suggestion
                                AMP_format_addon= "%%COMP(2**NPWAVE-1)",
```

madgraph/iolibs/export_v4.py
```diff
@@ -7313,6 +8264,9 @@ def copy_standard_file(self):
             load_card = ''
             lha_read_filename='lha_read.f'
             updateloop_default = '.false.'
+
+        cp( MG5DIR + '/models/template_files/fortran/rw_onia.f', \
+                                       os.path.join(self.dir_path,'rw_onia.f'))
```
oliviermattelaer, line 8278: > Since this file is part of the Source folder it shouldn't know anything about the specific process

Agreed -- so gate it on the **model**, not on the process. `setonia` can only ever be needed by a model that defines Fock states, and that is something the Source writer may legitimately know: `pass_information_from_cmd` already receives the interface.

Also worth noting: `rw_onia.f` is currently copied **twice** -- once here and once in `file_to_link` at line 8226 -- so the copy below is redundant today.

Four companion edits (all outside the PR diff, hence not offered as suggestions):

```python
# line 536, ProcessExporterFortran.pass_information_from_cmd
        self.has_boundstates = bool(getattr(cmd, '_fockstates', []))

# line 1385, ProcessExporterFortran.convert_model
        model_opts = self.opt + self.proc_characteristic
        model_opts['onia'] = getattr(self, 'has_boundstates', False)
        model_builder = UFO_model_to_mg4(model, write_dir, model_opts)

# line 7967, UFO_model_to_mg4.__init__ default
                        'loop_induced': False, 'onia': False}

# line 8226, drop the duplicate copy
        file_to_link = ['formats.inc', 'rw_para.f', 'testprog.f']

# line 10697, create_makeinc
        text = 'MODEL = couplings.o lha_read.o printout.o rw_para.o'
        if self.opt['onia']:
            text += ' rw_onia.o'
```

Verified: `p p > e+ e-` gets neither `rw_onia.f` nor `rw_onia.o` and `Source` builds clean; `g g > hc(1|1P11) g` (standalone and madevent) still gets both and links `check`. The three onia acceptance tests pass.

```suggestion
        if self.opt['onia']:
            cp( MG5DIR + '/models/template_files/fortran/rw_onia.f', \
                                       os.path.join(self.dir_path,'rw_onia.f'))
```

madgraph/iolibs/export_v4.py
LMaxia: That is smart; I actually thought that the % sign works on the left beside parentheses. Good to know!
I have updated the branch, removing our "_percent_" workaround.

madgraph/interface/madgraph_interface.py
lsimon125: I think it would make sense, but currently it breaks some tests if we include the 3FS consistently for non-onia models.

madgraph/various/banner.py
lsimon125: This is related to the question if it makes sense to implement the 3FS in a more consistent fashion. For all cases I checked this modification here was save and does what it should do. I am not sure if there is an edgecase where it could be broken. As I find the current definition of using a 4FS even when a charm quark is massive in the model is an unnatural assumption, setting the minium value of `maxjetflavor` to 3 sounds more reasonable.

madgraph/fks/fks_common.py
```diff
@@ -43,11 +43,11 @@ class FKSDiagramTag(diagram_generation.DiagramTag): #test written
...
         Note that the number is not taken into account if tag comparison,
         but is used only to extract leg permutations.
         """
-        return [((leg.get('id'), leg.get('number')), leg.get('number'))]
+        return [((leg.get('id'), leg.get('number'), leg.get('onium')), leg.get('number'))]
```
oliviermattelaer, line 50: No, it is not needed -- and it is slightly harmful.

`FKSDiagramTag` is only used on the FKS side, and an onium can never be initial state (`initial particles cannot be onia`) nor survive to NLO, so `leg.get('onium')` is always `{}` here. It only ever adds an empty dict to the tag tuple, which also makes that tuple unhashable/unorderable.

Reverting to the pre-PR form; 77 FKS unit tests pass with this.

```suggestion
        """Returns the default end link for a leg: ((id, number), number).
        Note that the number is not taken into account if tag comparison,
        but is used only to extract leg permutations.
        """
        return [((leg.get('id'), leg.get('number')), leg.get('number'))]
```

(The same is true of the `onium` entry in `DiagramTag.link_from_leg` in `diagram_generation.py` -- it sits in the *initial-state* branch, so it is always `{}` too. `CanonicalConfigTag` in `helas_objects.py` is different: there it is read back by `leg_from_link` and genuinely needed.)

madgraph/iolibs/export_v4.py
```diff
@@ -1252,7 +1616,7 @@ def get_color_data_lines(self, matrix_element, n=128):
         rows in chunks of size n."""
 
         if not matrix_element.get('color_matrix'):
-            return ["DATA %(proc_prefix)sDenom/1/", "DATA %(proc_prefix)sCF/1/"]
+            return ["DATA %(proc_prefix)sDenom/1/", "DATA %(proc_prefix)sCF/0/"]
```
oliviermattelaer: Applying your suggestion, adapted to the `%(proc_prefix)s` form that came in with the 3.8.0 merge.

Worth keeping the non-onia branch at `CF/1`: `ncolor` is forced to `max(1, len(color_basis))`, so with `CF/0` the generated code computes `MATRIX = 0` -- a silent zero rather than an error -- for any non-onia process that ever reaches this fallback.

```suggestion
            if matrix_element.get_nonia() > 0:
                # colour-forbidden Fock state: the matrix element has to vanish
                return ["DATA %(proc_prefix)sDenom/1/", "DATA %(proc_prefix)sCF/0/"]
            # unchanged for everything else: a single trivial colour structure
            return ["DATA %(proc_prefix)sDenom/1/", "DATA %(proc_prefix)sCF/1/"]
```

Template/LO/Cards/run_card.dat
oliviermattelaer: Splitting them as you asked.

For the record the side effect is real and not cosmetic: `RunCard.write` has two paths, and the line-by-line one (taken whenever any parameter is user-set, i.e. every rewrite of an existing card) parses the whole line as a single block name, gets `RUNNING$mom_resh`, logs `block ... not defined` and drops the marker. Both blocks are then re-appended at the very *end* of the card.

Measured on a card with one user-set parameter: `mue_over_ref` at line 268 and `mom_resh_type` at 278 before, 70 and 81 after.

```suggestion
$RUNNING
$mom_resh
```

madgraph/interface/madgraph_interface.py
```diff
@@ -5352,15 +5485,23 @@ def extract_process(self, line, proc_number = 0, overall_orders = {},
...
                     else:
                         myleglist.append(fks_tag.MultiTagLeg({'ids':mylegids,
                                                           'state':state,
                                                           'polarization': polarization,
+                                                          'onium': {},
                                                           'is_tagged':is_tagged}))
+            elif is_onium:
+                pass
             else:
                 raise self.InvalidCmd("No particle %s in model" % part_name)
```
oliviermattelaer, line 5495: > I found no guard refusing onia at NLO

Here is one. It sits right before the `aloha.dual_mode` bookkeeping in `extract_process`, where both `onium_index` and `perturbation_couplings` are known.

The hole is real: on the current branch `generate g g > hc(1|1P11) g [real=QCD]` is accepted and silently expands the onium into its `c c~` constituents -- the FKS/loop path never reads the `onium` leg properties back, and there is no onia matrix-element template for it. (`[QCD]` happens to be caught earlier by the model-level "cannot handle loop processes" check, but `[real=...]` and `[LOonly=...]` are not.)

Verified: the three forms above are now refused, plain `generate g g > hc(1|1P11) g` still works.

```suggestion

        # Bound states are handled by the LO exporters only: nothing in the
        # FKS/loop path reads back the 'onium' leg properties and there is no
        # onia matrix-element template for it, so a perturbed process would
        # silently produce a wrong result rather than fail.
        if onium_index and perturbation_couplings.strip():
            raise self.InvalidCmd(
                "Onia are only supported at leading order: the perturbation "
                "'[%s]' can not be combined with a bound state."
                % perturbation_couplings.strip())
```

madgraph/various/banner.py
```diff
@@ -4948,7 +4968,7 @@ def create_default_for_process(self, proc_characteristic, history, proc_def):
...
```
oliviermattelaer, line 4970: > setting the minimum value of `maxjetflavor` to 3 sounds more reasonable

I agree about the massive-charm case -- but the floor also decides the *opposite* case, and there it now gets it wrong.

With the **default sm** (charm massless, `p = g u c d s u~ c~ d~ s~`, i.e. 4FS), `generate g g > t t~` has no quark in `beam_id`, so `max([3])` gives `maxjetflavor = 3` against a four-flavour jet definition. It was 4 before this PR. That is a plain-SM regression, not an onia corner.

Deriving the floor from the model gives you what you want in both directions -- 3 exactly when charm is massive, 4 when it is not:

| model | `j`/`p` | before | after |
|---|---|---|---|
| `sm` (`g g > t t~`) | 4FS | 3 | **4** |
| `sm` (`p p > t t~`) | 4FS | 4 | 4 |
| `sm-no_b_mass` | 5FS | 5 | 5 |
| `sm-c_mass` | 3FS | 3 | 3 |
| `sm_onia-c_mass` | 3FS | 3 | 3 |

```suggestion
                # The floor has to follow the flavour scheme that
                # add_default_multiparticles picked for j/p, i.e. the heaviest
                # quark that is still massless in this model. Hard-coding it
                # makes a gluon-initiated process disagree with its own jet
                # definition.
                nflav = 3
                try:
                    model = proc_def[0][0].get('model')
                    massless = [pdg for pdg in range(1, 6)
                                if model.get_particle(pdg) and
                                   model.get_particle(pdg).get('mass') == 'ZERO']
                    if massless:
                        nflav = max(massless)
                except Exception:
                    pass
                maxjetflavor = max([nflav]+[abs(i) for i in beam_id if  -7< i < 7])
```

madgraph/interface/madgraph_interface.py
```diff
@@ -6164,14 +6309,30 @@ def add_default_multiparticles(self):
...
                 break
+            if not c:
+                break
 
+            if 4 in multi:
+                if c['mass'] != 'ZERO':
+                    if 'onia' in self._curr_model.get('name'):
+                        multi.remove(4)
+                        multi.remove(-4)
+                        scheme = 3
```
oliviermattelaer, line 6329: > currently it breaks some tests if we include the 3FS consistently for non-onia models

I tried to reproduce that and could not. Full acceptance `test_cmd` suite (42 tests), three runs:

```
with the change applied:            2 failures, 8 errors
with only this hunk reverted:       2 failures, 8 errors
clean branch, no local changes:     2 failures, 8 errors
```

The ten are all MadSpin density-matrix tests failing on a missing `numpy` in my environment (`test_density_mode_*`, `test_standalone_density_*`, and `test_madevent_subproc_group`, whose sibling names the missing `Average_density_matrix_unweighted_events.txt`). The full unit suite is likewise unchanged by this hunk.

Could you say which tests you saw break? My guess is the parallel `test_ML5` c-massive comparisons, which need external runners I do not have here.

The argument for making it unconditional is that the `b` branch immediately below already does exactly this without a model-name test, so today a massive `c` stays in `j`/`p` while a massive `b` is removed.

```suggestion
                # a massive c has no business in the massless-jet definition,
                # exactly as for the b just below: do it for every model, not
                # only for the onia ones.
                if c['mass'] != 'ZERO':
                    multi.remove(4)
                    multi.remove(-4)
                    scheme = 3
```

madgraph/interface/madgraph_interface.py
```diff
@@ -9859,14 +10050,24 @@ def finalize(self, nojpeg, online = False, flaglist=[]):
...
-                                               wanted_couplings)
+                for npwave in aloha.npwave:
+                    if npwave == 0:
+                        aloha.dual_mode = False
+                    else:
+                        aloha.dual_mode = True
+                    self._curr_exporter.convert_model(self._curr_model,
+                                                   wanted_lorentz,
+                                                   wanted_couplings,
+                                                   npwave)
```
oliviermattelaer: Taking your second option -- `**kwargs` at the end of every exporter, so an option meant for one of them is simply ignored by the others. `npwave` is then passed by keyword and only `ProcessExporterFortran` names it.

This is 1/10 of the change; the other nine are one-liners posted on their own lines so you can apply them individually. All ten go together.

```suggestion
                                                   npwave=npwave)
```

madgraph/iolibs/export_v4.py
```diff
@@ -147,7 +147,7 @@ def generate_subprocess_directory_end(self, **opt):
         """
         return 
 
-    def convert_model(self, model, wanted_lorentz=[], wanted_couplings=[]):
+    def convert_model(self, model, wanted_lorentz=[], wanted_couplings=[], npwave=0):
```
oliviermattelaer: 2/10 of the `**opts` change -- VirtualExporter stub.

```suggestion
    def convert_model(self, model, wanted_lorentz=[], wanted_couplings=[], **opts):
```

madgraph/iolibs/export_v4.py
```diff
@@ -1021,7 +1366,8 @@ def write_maxamps_file(self, writer, maxamps, maxflows,
     #===========================================================================
 
     def convert_model(self, model, wanted_lorentz = [],
-                             wanted_couplings = []):
+                             wanted_couplings = [],
+                             npwave = 0):
         """ Create a full valid MG4 model from a MG5 model (coming from UFO)"""
```
oliviermattelaer: 3/10 of the `**opts` change -- the one exporter that actually uses npwave.

```suggestion
                             npwave = 0, **opts):
        """ Create a full valid MG4 model from a MG5 model (coming from UFO)

        Every exporter takes **opts so that an option meant for one of them
        (npwave here) can be passed by keyword to all of them and silently
        ignored by those that do not care about it."""
```

madgraph/iolibs/export_v4.py
```diff
@@ -3562,10 +4282,10 @@ def handle_cuts_inc(self):
     # convert_model
     #===========================================================================    
     def convert_model(self, model, wanted_lorentz = [], 
-                                                         wanted_couplings = []):
+                             wanted_couplings = [], npwave = 0):
```
oliviermattelaer: 4/10 of the `**opts` change -- ProcessExporterFortranMW.

```suggestion
                             wanted_couplings = [], **opts):
```

madgraph/iolibs/export_v4.py
```diff
@@ -3562,10 +4282,10 @@ def handle_cuts_inc(self):
...
          
         super(ProcessExporterFortranMW,self).convert_model(model, 
-                                               wanted_lorentz, wanted_couplings)
+                                               wanted_lorentz, wanted_couplings, npwave)
```
oliviermattelaer: 5/10 of the `**opts` change -- ...and its super() call.

```suggestion
                                               wanted_lorentz, wanted_couplings, **opts)
```

madgraph/iolibs/export_v4.py
```diff
@@ -4451,10 +5173,10 @@ def copy_python_file(self):
  
  
     def convert_model(self, model, wanted_lorentz = [], 
-                                                         wanted_couplings = []):
+                            wanted_couplings = [], npwave = 0):
```
oliviermattelaer: 6/10 of the `**opts` change -- ProcessExporterFortranME.

```suggestion
                            wanted_couplings = [], **opts):
```

madgraph/iolibs/export_v4.py
```diff
@@ -4451,10 +5173,10 @@ def copy_python_file(self):
...
          
         super(ProcessExporterFortranME,self).convert_model(model, 
-                                               wanted_lorentz, wanted_couplings)
+                                               wanted_lorentz, wanted_couplings, npwave)
```
oliviermattelaer: 7/10 of the `**opts` change -- ...and its super() call.

```suggestion
                                               wanted_lorentz, wanted_couplings, **opts)
```

madgraph/iolibs/export_cpp.py
```diff
@@ -2634,7 +2634,7 @@ def modify_grouping(self, matrix_element):
 
 
     def convert_model(self, model, wanted_lorentz = [],
-                         wanted_couplings = []):
+                         wanted_couplings = [], npwave = 0):
```
oliviermattelaer: 8/10 of the `**opts` change -- ProcessExporterCPP.

```suggestion
                         wanted_couplings = [], **opts):
```

madgraph/loop/loop_exporters.py
```diff
@@ -507,7 +507,7 @@ def write_loop_makefile_definitions(self, writer, link_tir_libs,
             return file
         
     def convert_model(self, model, wanted_lorentz = [], 
-                                                         wanted_couplings = []):
+                                             wanted_couplings = [], npwave = 0):
```
oliviermattelaer: 9/10 of the `**opts` change -- LoopProcessExporterFortranSA.

```suggestion
                                             wanted_couplings = [], **opts):
```

madgraph/iolibs/export_fks.py
```diff
@@ -334,7 +334,7 @@ def copy_python_files(self):
                 pjoin(self.dir_path,'bin','internal',os.path.basename(cp_file)))
 
     def convert_model(self, model, wanted_lorentz = [], 
-                                                         wanted_couplings = []):
+                                            wanted_couplings = [], npwave = 0):
```
oliviermattelaer: 10/10 of the `**opts` change -- ProcessExporterFortranFKS (keeps deliberately not forwarding npwave).

```suggestion
                                            wanted_couplings = [], **opts):

        # npwave (if any) is deliberately not forwarded: the FKS output has no
        # dual-number HELAS library.
```

aloha/__init__.py
```diff
@@ -5,6 +5,7 @@
                       # 2: axial
                       # 3: Feynman Diagram gauge (5D aloha)
 loop_mode = False     # Tag for encoding momenta with complex number.
+dual_mode = False     # Tag for encoding momenta with dual number.
```
oliviermattelaer: > Declare `npwave = [0]` in `aloha/__init__.py` and drop the `hasattr` guards ... and the variable is typed three ways

Both, in one place. `npwave` becomes a real module global (so it is discoverable and the `hasattr` dance goes away), and `dual_mode` becomes consistently an int count -- `0` means off, which keeps every `if aloha.dual_mode:` consumer working unchanged while making `npwave.append(aloha.dual_mode)` and `dual_mode += 1` mean the same thing as the declaration.

1/6; the five companions follow (two `hasattr` drops, two `dual_mode` assignments, and `set_global`).

```suggestion
dual_mode = 0         # Number of P-wave onia in the current process (0/False
                      # when none): selects the dual-number HELAS routines.
npwave = [0]          # P-wave multiplicities the exporter has to write HELAS
                      # libraries for. 0 is the plain (non-dual) library.
```

madgraph/interface/madgraph_interface.py
```diff
@@ -4860,6 +4881,10 @@ def do_generate(self, line):
         self.clean_process()
         self._generate_info = line
 
+        # reset P-wave onia counter for new process
+        if hasattr(aloha, "npwave"):
+            aloha.npwave = [0]
```
oliviermattelaer: 2/6 -- `npwave` is a declared global now, so the guard is dead.

```suggestion
        aloha.npwave = [0]
```

madgraph/interface/madgraph_interface.py
```diff
@@ -5155,6 +5180,57 @@ def extract_process(self, line, proc_number = 0, overall_orders = {},
...
+        boundstates_keys = [key for key in self._boundstates.keys()]
+        boundstates_keys_lower = [key.lower() for key in boundstates_keys]
+        for index,part_name in enumerate(args):
+            if part_name in boundstates_keys_lower:
+                bound_name = boundstates_keys[boundstates_keys_lower.index(part_name)]
+                boundstates[index] = self._boundstates[bound_name]
+
+        if not hasattr(aloha, "npwave"):
+            aloha.npwave = [0]
+
```
oliviermattelaer: 3/6 -- the second `hasattr` guard, likewise dead. (Deleting these two is what removes the cross-command state dependency: whether `npwave` existed used to depend on which command had run before.)

```suggestion
```

madgraph/interface/madgraph_interface.py
```diff
@@ -5164,6 +5240,8 @@ def extract_process(self, line, proc_number = 0, overall_orders = {},
         else:
             upc_with_jet = False
 
+        onium_index = 0
+        aloha.dual_mode = False
```
oliviermattelaer: 4/6 -- `dual_mode` is a counter from here on (`+= 1` per P-wave onium just below), so start it at `0` rather than `False`.

```suggestion
        aloha.dual_mode = 0
```

madgraph/interface/madgraph_interface.py
```diff
@@ -9859,14 +10050,24 @@ def finalize(self, nojpeg, online = False, flaglist=[]):
...
                 self.previous_couplings = wanted_couplings
             else:
-                self._curr_exporter.convert_model(self._curr_model, 
-                                               wanted_lorentz,
-                                               wanted_couplings)
+                for npwave in aloha.npwave:
+                    if npwave == 0:
+                        aloha.dual_mode = False
+                    else:
+                        aloha.dual_mode = True
```
oliviermattelaer: 5/6 -- with `dual_mode` typed as a count, the loop body is just the count itself.

```suggestion
                    aloha.dual_mode = npwave
```

madgraph/iolibs/export_v4.py
```diff
@@ -6373,7 +7268,18 @@ def generate_subprocess_directory(self, subproc_group,
 
         matrix_elements = subproc_group.get('matrix_elements')
 
+        contains_onia = False
+        if matrix_elements[0].get_nonia()>0:
+            if matrix_elements[0].get_npwave()>0:
+                aloha.dual_mode = True
+            else:
+                aloha.dual_mode = False
```
oliviermattelaer: 6/6 -- same here: assign the P-wave count instead of a bool.

```suggestion
            aloha.dual_mode = matrix_elements[0].get_npwave()
```

madgraph/iolibs/export_v4.py
```diff
@@ -1039,8 +1385,9 @@ def convert_model(self, model, wanted_lorentz = [],
         model_builder = UFO_model_to_mg4(model, write_dir, self.opt + self.proc_characteristic)
```
oliviermattelaer: Companion to the `rw_onia.f` gating comment above (gate on the *model* having Fock states, which the Source writer may legitimately know). 2/4 -- hand the flag down to the model writer.

```suggestion
        model_opts = self.opt + self.proc_characteristic
        model_opts['onia'] = getattr(self, 'has_boundstates', False)
        model_builder = UFO_model_to_mg4(model, write_dir, model_opts)
```

madgraph/iolibs/export_v4.py
```diff
@@ -7272,7 +8223,7 @@ def copy_standard_file(self):
         """Copy the standard files for the fortran model."""
         
         #copy the library files
-        file_to_link = ['formats.inc', \
+        file_to_link = ['formats.inc', 'rw_onia.f', \
                         'rw_para.f', 'testprog.f']
```
oliviermattelaer: Companion to the `rw_onia.f` gating comment above (gate on the *model* having Fock states, which the Source writer may legitimately know). 3/4 -- `rw_onia.f` is currently copied twice; this is the duplicate. The gated copy further down is the one to keep.

```suggestion
        # rw_onia.f is copied further down, only for a model with bound states
        file_to_link = ['formats.inc', 'rw_para.f', 'testprog.f']
```

madgraph/iolibs/export_v4.py
```diff
@@ -9740,7 +10694,7 @@ def create_makeinc(self):
         """create makeinc.inc containing the file to compile """
         
         fsock = self.open('makeinc.inc', comment='#')
-        text = 'MODEL = couplings.o lha_read.o printout.o rw_para.o'
+        text = 'MODEL = couplings.o lha_read.o printout.o rw_onia.o rw_para.o'
```
oliviermattelaer: Companion to the `rw_onia.f` gating comment above (gate on the *model* having Fock states, which the Source writer may legitimately know). 4/4 -- and keep `rw_onia.o` out of the model library when it is not needed.

```suggestion
        text = 'MODEL = couplings.o lha_read.o printout.o rw_para.o'
        if self.opt['onia']:
            text += ' rw_onia.o'
```

The two remaining edits (`pass_information_from_cmd` at line 536 and the `UFO_model_to_mg4` default at 7967) are outside the PR diff, so GitHub will not take them as suggestions -- they are in the parent comment.

oliviermattelaer:
### Remaining review points -- patches for the bits GitHub cannot take as suggestions

I have put inline `suggestion` blocks on everything that lives inside the PR diff (fks_common, the `CF/0` fallback, the run_card block split, the NLO guard, `maxjetflavor`, the 3FS multiparticles, `**opts` on the exporters 1/10..10/10, the aloha globals 1/6..6/6, and the `rw_onia` companions 2/4..4/4).

Three of your points touch lines or files that are **not** part of this PR's diff, so GitHub refuses an inline suggestion on them. The diffs are below -- they apply cleanly on `db3daf569`.

---

#### 1. `detect_card_type` has no onia entry, and `reload_card` no onia branch

`detect_card_type` returning `unknown` for an onia card is not harmless: `AskforEditCard.copy_file` warns and then falls through to `self.paths['unknown']`, i.e. `KeyError`. The token has to be `block\s+ldme` rather than bare `ldme`, otherwise the extended `ident_card.dat` (which gets plain `ldme <id> LDME_<id>` lines) is caught too.

`reload_card` matters because `do_set` already re-reads `self.onia_card` explicitly, so it is clearly meant to stay in sync -- without the branch an edit made in the editor is silently ignored.

Checked against nine card types (param/run/plot/delphes/pgs/madspin/rivet/onia/restrict) with no misdetection, and confirmed an external edit is now picked up.

```diff
diff --git a/madgraph/interface/common_run_interface.py b/madgraph/interface/common_run_interface.py
index 4626d437f..9c77958e3 100755
--- a/madgraph/interface/common_run_interface.py
+++ b/madgraph/interface/common_run_interface.py
@@ -1199,6 +1199,7 @@ class CommonRunCmd(HelpToCmd, CheckValidForCmd, cmd.Cmd):
            madanalysis5_hadron_card.dat
            madanalysis5_parton_card.dat
            rivet_card.dat
+           onia_card.dat
            
            Please update the unit-test: test_card_type_recognition when adding
            cards.
@@ -1227,6 +1228,7 @@ class CommonRunCmd(HelpToCmd, CheckValidForCmd, cmd.Cmd):
                     'gridpack',
                     'ebeam1',
                     r'block\s+mw_run',
+                    r'block\s+ldme',
                     'BLOCK',
                     'DECAY',
                     'launch',
@@ -1279,6 +1281,8 @@ class CommonRunCmd(HelpToCmd, CheckValidForCmd, cmd.Cmd):
             return 'madweight_card.dat'
         elif 'transfer_card.dat' in text:
             return 'transfer_card.dat'
+        elif any(t.endswith('ldme') for t in text):
+            return 'onia_card.dat'
         elif 'block' in text and 'decay' in text: 
             return 'param_card.dat'
         elif 'b_stable' in text:
@@ -8332,6 +8336,14 @@ You can also copy/paste, your event file here.''')
                 self.run_card = banner_mod.RunCard(path)
         elif path == self.paths['shower']:
             self.shower_card = shower_card_mod.ShowerCard(path)
+        elif path == self.paths['onia']:
+            try:
+                self.onia_card = param_card_mod.ParamCard(path)
+            except (param_card_mod.InvalidParamCard, ValueError) as e:
+                logger.error('Current onia_card is not valid.')
+                logger.error('problem detected: %s' % e)
+                logger.error('Please re-open the file and fix the problem.')
+                logger.warning('using the \'set\' command without opening the file will discard all your manual change')
         elif path == self.paths['ML']:
             self.MLcard = banner_mod.MadLoopParam(path)
         elif path == self.paths['pythia8']:
```

#### 2. `set_global` does not reset the new globals

Both copies of the decorator, plus the unit-test entry that the `detect_card_type` docstring asks for ("Please update the unit-test: test_card_type_recognition when adding cards").

`npwave` is a list mutated in place, so it needs a copy rather than a reference. Verified that both globals are restored after a mutation *and* after an exception, and that the stored list is not aliased.

```diff
diff --git a/madgraph/various/misc.py b/madgraph/various/misc.py
index c6da20700..310a8060e 100755
--- a/madgraph/various/misc.py
+++ b/madgraph/various/misc.py
@@ -2183,7 +2183,8 @@ It has been validated for the last time with version: %s""",
     
 
 #decorator
-def set_global(loop=False, unitary=True, mp=False, cms=False):
+def set_global(loop=False, unitary=True, mp=False, cms=False,
+               dual=0, npwave=(0,)):
     from functools import wraps
     import aloha
     import aloha.aloha_lib as aloha_lib
@@ -2194,10 +2195,15 @@ def set_global(loop=False, unitary=True, mp=False, cms=False):
             old_gauge = aloha.unitary_gauge
             old_mp = aloha.mp_precision
             old_cms = aloha.complex_mass
+            old_dual = aloha.dual_mode
+            # npwave is a list mutated in place, so keep a copy of it
+            old_npwave = list(aloha.npwave)
             aloha.loop_mode = loop
             aloha.unitary_gauge = unitary
             aloha.mp_precision = mp
             aloha.complex_mass = cms
+            aloha.dual_mode = dual
+            aloha.npwave = list(npwave)
             aloha_lib.KERNEL.clean()
             try:
                 out =  f(*args, **opt)
@@ -2206,11 +2212,15 @@ def set_global(loop=False, unitary=True, mp=False, cms=False):
                 aloha.unitary_gauge = old_gauge
                 aloha.mp_precision = old_mp
                 aloha.complex_mass = old_cms
+                aloha.dual_mode = old_dual
+                aloha.npwave = old_npwave
                 raise
             aloha.loop_mode = old_loop
             aloha.unitary_gauge = old_gauge
             aloha.mp_precision = old_mp
             aloha.complex_mass = old_cms
+            aloha.dual_mode = old_dual
+            aloha.npwave = old_npwave
             aloha_lib.KERNEL.clean()
             return out
         return deco_f_set
diff --git a/tests/IOTests.py b/tests/IOTests.py
index 3bd11f94b..a27dd82b0 100755
--- a/tests/IOTests.py
+++ b/tests/IOTests.py
@@ -55,7 +55,8 @@ _hc_comparison_files = pjoin(_input_file_path,'IOTestsComparison')
 _hc_comparison_tarball = pjoin(_input_file_path,'IOTestsComparison.tar.bz2')
 
 
-def set_global(loop=False, unitary=True, mp=False, cms=False):
+def set_global(loop=False, unitary=True, mp=False, cms=False,
+               dual=0, npwave=(0,)):
     """This decorator set_global() which make sure that for each test
     the global variable are returned to their default value. This decorator can
     be modified with the new global variables to come and will potenitally be
@@ -67,10 +68,15 @@ def set_global(loop=False, unitary=True, mp=False, cms=False):
             old_gauge = aloha.unitary_gauge
             old_mp = aloha.mp_precision
             old_cms = aloha.complex_mass
+            old_dual = aloha.dual_mode
+            # npwave is a list mutated in place, so keep a copy of it
+            old_npwave = list(aloha.npwave)
             aloha.loop_mode = loop
             aloha.unitary_gauge = unitary
             aloha.mp_precision = mp
             aloha.complex_mass = cms
+            aloha.dual_mode = dual
+            aloha.npwave = list(npwave)
             aloha_lib.KERNEL.clean()
             try:
                 out =  f(*args, **opt)
@@ -79,11 +85,15 @@ def set_global(loop=False, unitary=True, mp=False, cms=False):
                 aloha.unitary_gauge = old_gauge
                 aloha.mp_precision = old_mp
                 aloha.complex_mass = old_cms
+                aloha.dual_mode = old_dual
+                aloha.npwave = old_npwave
                 raise
             aloha.loop_mode = old_loop
             aloha.unitary_gauge = old_gauge
             aloha.mp_precision = old_mp
             aloha.complex_mass = old_cms
+            aloha.dual_mode = old_dual
+            aloha.npwave = old_npwave
             aloha_lib.KERNEL.clean()
             return out
         return deco_f_set
diff --git a/tests/unit_tests/interface/test_madevent.py b/tests/unit_tests/interface/test_madevent.py
index c9f44b806..81465816d 100755
--- a/tests/unit_tests/interface/test_madevent.py
+++ b/tests/unit_tests/interface/test_madevent.py
@@ -119,6 +119,14 @@ class TestMadEventCmd(unittest.TestCase):
         card_dir= pjoin(root_path,'..','Template', 'LO', 'Cards')
         self.assertEqual(detect(pjoin(card_dir, 'rivet_card_default.dat')),
                          'rivet_card.dat')        
+
+        # Onia card
+        self.assertEqual(detect(pjoin(root_path, 'input_files',
+                                      'onia_card_quarkonium.dat')),
+                         'onia_card.dat')
+        self.assertEqual(detect(pjoin(root_path, 'input_files',
+                                      'onia_card_leptonium.dat')),
+                         'onia_card.dat')
         
         
     def test_help_category(self):
```

#### 3. Two `rw_onia` companions outside the diff

Referenced from the `rw_onia` suggestions 2/4..4/4.

```diff
@@ -534,6 +534,9 @@ C
         """Pass information for MA5"""
         
         self.proc_defs = cmd._curr_proc_defs
+        # only a model that defines Fock states can ever need setonia (and
+        # therefore rw_onia.f) in the model library
+        self.has_boundstates = bool(getattr(cmd, '_fockstates', []))
 
     #===========================================================================
     # Create jpeg diagrams, html pages,proc_card_mg5.dat and madevent.tar.gz
@@ -7964,7 +7974,7 @@ class UFO_model_to_mg4(object):
         self.dir_path = output_path
         
         self.opt = {'complex_mass': False, 'export_format': 'madevent', 'mp':True,
-                        'loop_induced': False}
+                        'loop_induced': False, 'onia': False}
         if opt:
             self.opt.update(opt)
```

---

With all of the above plus the inline suggestions applied, on my side: full unit suite unchanged from the branch baseline, acceptance `test_cmd` unchanged, and `test_quarkonium_standalone` / `test_quarkonium_production` / `test_leptonium_production` all pass.

madgraph/interface/madgraph_interface.py
lsimon125: The test that was failing was `test_import_from_cmd`. But this is a long while ago and there were some other issues that caused many tests to fail that actually should have worked. So if this test works well with your changes I would recommend to use the 3 flavour scheme consistently. And apply your suggestion.

oliviermattelaer:
### :warning: `onia_pwave` is currently broken for onia output -- the `rw_onia` gating is half applied

Two of the four `rw_onia` suggestions landed (`model_opts['onia'] = ...` and the `create_makeinc` gate) but the two that *set* the flag did not, so `self.opt['onia']` is always `False`:

* `pass_information_from_cmd` never sets `has_boundstates` -- the name occurs exactly once in the file, in the `getattr(self, 'has_boundstates', False)` that reads it
* `UFO_model_to_mg4.__init__` has no `'onia'` default

Net effect: `rw_onia.o` is now gated **out** of `makeinc.inc` for every process, while `rw_onia.f` is still copied. `setonia` is therefore never compiled.

On `a27dc7ac` :

```
$ import model sm_onia-c_mass ; generate g g > hc(1|1P11) g ; output standalone
$ cd Source && make && cd ../SubProcesses/P1_gg_hc11P11g && make
Undefined symbols for architecture arm64:
  "_setonia_", referenced from:
      _MAIN__ in check_sa.o
ld: symbol(s) not found for architecture arm64
```

Applying the two missing pieces (they are in [this comment](https://github.com/mg5amcnlo/mg5amcnlo/pull/332#issuecomment-5620444232), section 3) restores it -- `rw_onia.o` back in `makeinc.inc`, `check` links, onia acceptance tests pass.

Sorry -- this is the partial-application hazard I flagged on those suggestion sets; the two pieces that could not be offered as suggestions are exactly the ones that got missed. Also still unapplied is suggestion 3/4 (dropping the duplicate `rw_onia.f` copy in `file_to_link`), which is cosmetic by comparison.

Same caveat applies to the `**opts` set (1/10..10/10) and the aloha set (1/6..6/6): I see the accidental deletions in `39a0801` and `b7fe32d` were caught and repaired -- thanks, and sorry for the noise. Those two were my fault: I anchored an insert on the following line instead of the preceding one, and two of my suggested ranges overlapped.

madgraph/interface/madgraph_interface.py
oliviermattelaer: You were right and I was wrong -- apologies. `test_import_from_cmd` **does** break with the 3FS change, and my earlier "cannot reproduce" was bad measurement on my side: at the time that test was already red on this branch for an unrelated environment reason, so I wrote it off as pre-existing. Re-checked cleanly just now:

```
clean a27dc7ac:                    OK
a27dc7ac + the 3FS change:         FAILED
```

The cause is a latent bug that the change merely exposes. `test_import_from_cmd` imports `231_Model_UFO`, which defines a real particle called `j`, so the `j` *multiparticle* is deleted at the top of `add_default_multiparticles`. But the re-registration at the end sits **outside** the `for qcd_container in ['p','j']` loop:

```python
        if scheme in [3,4,5] and not photon:
            self.optimize_order(multi)
            self._multiparticles[qcd_container] = multi
```

`qcd_container` is whatever the loop left behind (`'j'`), so as soon as `scheme` is set the deleted `j` is silently resurrected. Today `scheme` only gets set for onia models, which is why nobody hit it.

Guarding it fixes the test and keeps everything else identical:

```suggestion
        if scheme in [3,4,5] and not photon:
            self.optimize_order(multi)
            # only re-register the containers that still exist: a model may
            # define a real particle called 'j' (or 'p'), in which case the
            # multiparticle was dropped above and must not come back.
            for container in ['p', 'j']:
                if container in self._multiparticles:
                    self._multiparticles[container] = multi
```

With this plus the 3FS suggestion: `test_import_from_cmd` OK, whole `test_import_ufo` module OK (35 tests), and the schemes still come out right -- `sm` 4FS/`maxjetflavor` 4, `sm-c_mass` 3FS/3, `sm-no_b_mass` 5FS/5, `sm_onia-c_mass` 3FS/3.

madgraph/interface/madgraph_interface.py
```diff
@@ -6187,7 +6354,7 @@ def add_default_multiparticles(self):
                 multi.remove(22)
                 photon = False
                 
-        if scheme in [4,5] and not photon:
+        if scheme in [3,4,5] and not photon:
             self.optimize_order(multi)
             self._multiparticles[qcd_container] = multi
```
oliviermattelaer, line 6359: Companion to the 3FS change -- this is the line that actually breaks `test_import_from_cmd`.

The re-registration is outside the `for qcd_container in ['p','j']` loop, so `qcd_container` is whatever the loop left behind (`'j'`). When a model defines a real particle called `j`, the `j` multiparticle is deleted at the top of this method and then silently resurrected here as soon as `scheme` is set. Only onia models set `scheme` today, which is why it has stayed hidden.

```suggestion
        if scheme in [3,4,5] and not photon:
            self.optimize_order(multi)
            # only re-register the containers that still exist: a model may
            # define a real particle called 'j' (or 'p'), in which case the
            # multiparticle was dropped above and must not come back.
            for container in ['p', 'j']:
                if container in self._multiparticles:
                    self._multiparticles[container] = multi
```

oliviermattelaer:
One more issue:

import model sm              ->   0 fock states
import model sm_onia-c_mass  -> 192 fock states
import model sm              -> 192 fock states   

So we need to remove previous fock states when reloading model
I can handle detect_card_type, reload_card, both set_global if you do not mind.
And maybe handle myself the 3/4 flavor as well (We need to be carefull on that one).

But otherwise we seem very close to merge this :-D

lsimon125:
> One more issue:
> 
> import model sm -> 0 fock states import model sm_onia-c_mass -> 192 fock states import model sm -> 192 fock states
> 
> So we need to remove previous fock states when reloading model I can handle detect_card_type, reload_card, both set_global if you do not mind. And maybe handle myself the 3/4 flavor as well (We need to be carefull on that one).
> 
> But otherwise we seem very close to merge this :-D

Thank you very much for taking care it. I am fine with the last changes you suggested, they can be applied. The solution for 3/4 flavor scheme I also saw, I just wasn't confident enough to apply them as I didn't know if the original version was on purpose missing this redefinition of the proton and the jet.
Regarding the issue with the Fock state after loading another model: Even though a bit fishy, this can be used as feature, i.e. if first the model `sm_onia` is loaded and then another model it's possible to use bound states in the new model.  The corresponding bound states are indeed treated consistently in the last imported model, I have used this trick for example to compute some bound states in HEFT by first using `import sm_onia` and then `import heft`. If you wish, I can reset the Fock states when a new model gets imported, but this loophole can be handy, which is why I kept it on purpose.

oliviermattelaer:
So I have change the import model to clean the fock state.
But I have added an option to the add model option:
import model heft
add model sm_onia --fock_states_only
which lead to the same model (but in an explicit way).

If you agree with those changes (and if the CI is green) I will merge it.
Thanks a lot,

Olivier

oliviermattelaer (approved):

oliviermattelaer:
MERGED!!!!

lsimon125:
> So I have change the import model to clean the fock state. But I have added an option to the add model option: import model heft add model sm_onia --fock_states_only which lead to the same model (but in an explicit way).
> 
> If you agree with those changes (and if the CI is green) I will merge it. Thanks a lot,
> 
> Olivier

Thanks, a lot. This is a nice solution.

lsimon125:
> MERGED!!!!

Thank you! :+1:

PR #334, opened 2026-08-14
MadSpin sequential: restore the offshell rate factor in the mass-set …

oliviermattelaer:
…weight

The offshell (madspin/full) sequential accept/reject unweights in two stages: the mass set first, then each slot's decay angles. The per-angle stage redraws until it accepts and so divides out its own normalisation

    Z_k(m) = Int dPhi_off(m) |M_dec|^2 / Int dPhi_on |M_dec|^2 = (m/M) Gamma(m)/Gamma(M)

which is a strong function of the sampled virtuality. The accepted mass sets were therefore missing prod_k Z_k(m_k) and the reconstructed resonance lineshape came out Breit-Wigner (PA) shaped instead of offshell: on p p > t t~, t > w+ b, w+ > l+ vl, <m(l+ v b)> sat -0.248 GeV (-7.8 sigma) from the joint accept/reject, chi2/ndf 75.8/22 over a +-8 GeV window.

Z_k is a function of that slot's virtuality alone -- the production event, the other slots' masses and the angles already accepted all cancel out of it -- so it is tabulated per slot and multiplied into the mass-set weight. The samples are free: the max-weight probe already draws Nevents_for_max_weight * max_weight_ps_point chains, each giving one (m_k, s_k) pair per slot with s_k = jac_dec * Tr(D^off)/|M_dec|^2_on, which has the same conditional expectation as w_k without its polarisation modulation. The table is a weighted quadratic in ln(m/pole) through binned means (Z_k is an expectation, so the mean estimates it and the mean of the logarithms would not), and C_mass is derived afterwards from the completed weights -- hence the probe now keeps its chains instead of maxing them online.

An imperfect Z_hat does not cancel: the per-angle stage divides out the true Z_k whatever weight it is given, so the residual bias is exactly Z_hat/Z. Hence also the new sequential_exact option, where the mass stage pays Z_hat, each slot divides it back out and a rejected decay trashes the mass set. The chain is then accepted with probability proportional to the joint weight, so Z_hat cancels identically and is reduced to an efficiency preconditioner -- exact whatever the table says, at roughly 3x the cost.

Also: a decay that cannot be reshuffled onto its virtuality (jac_dec = 0) is now an ordinary rejection of that decay rather than a restart of the whole chain. The restart was a second, C_k-dependent mass-dependent normalisation (the set survived slot k with probability Z_k/(Z_k + q_k C_k), not Z_k) which would have defeated the correction near a decay threshold; it is invisible on t > W b, where m_t > M_W + M_b always. Counting those zeros as rejections is what makes the tabulated Z_k the exact correction there.

PA/onshell are untouched: every new line is inside an offshell branch, and neither the random-number stream nor any arithmetic on that path changes.

Validation (10000 events, seed 42, nb_core 1, same production sample):
- the fitted table reproduces the narrow-width (m/M) Gamma(m)/Gamma(M) to under 0.5%: Z(150.7) = 0.53, Z(173) = 1, Z(195.4) = 1.71;
- four sequential replicas with independent MadSpin seeds average both resonances to 173.1641 +- 0.0177 against two joint replicas at 173.1853 and 173.1867 -- a residual of -0.022 +- 0.024 GeV, consistent with zero;
- lineshape chi2/ndf 19.0/22, against 21.7/22 for a joint-vs-joint control;
- sequential_exact lands at +0.006 +- 0.032 GeV, chi2/ndf 10.4/22.

Cost, measured on the same runs: the offshell sequential path is ~2x slower than the joint test (28.7 s vs 14.4 s of decay-phase wall time), which the earlier per-event decay-ME count missed -- the mass-set stage draws 26 virtuality sets per accepted event, each an Event parse, a production reshuffle and a production density. Z_hat accounts for about a third of that gap (C_mass 14 -> 17-19). So sequential_decay = auto keeps routing madspin/full to the joint accept/reject; what this buys is that the offshell path is correct when switched on explicitly.

The decay-reshuffling jacobian this builds on (_decay_reshuffle_jacobian and the jac_dec factors in both branches) was already in the working tree, uncommitted, and is included here.

## Summary by Sourcery

Tabulate and apply per-slot offshell decay rate factors in MadSpin's sequential offshell unweighting, add an exact sequential mode, and adjust handling, statistics, caching, and tests accordingly to restore the correct resonance lineshape and provide validation of the scheme.

Enhancements:
- Introduce an offshell rate-factor tabulation pipeline (Z_k tables) from max-weight probe samples and use it to correct sequential mass-set weights in offshell spin modes.
- Add a sequential_exact option that rejects entire mass sets on decay rejection so the scheme is exact regardless of Z_k accuracy, and adjust logging and statistics for mass-stage performance and infeasible decays.
- Refine offshell sequential decay handling by treating infeasible decay reshuffles as local rejections, propagating decay reshuffling Jacobians into weights, and sharing BW and decay Jacobians consistently with the joint scheme.
- Extend max-weight scanning and caching to handle offshell sequential data, including storing Z_k tables alongside slot bounds and completing offshell probes post-fit.

Documentation:
- Expand MADSPIN_SEQUENTIAL_PLAN.md with a detailed discussion of the missing offshell normalisation issue, Z_k derivation and implementation, the sequential_exact mode, performance measurements, and updated validation results.

Tests:
- Add unit tests for Z_k(tabulated offshell rate factor) construction, interpolation, threshold handling, and its integration into probe completion, and for a toy model demonstrating bias removal and correctness of the sequential_exact mode.
- Update existing sequential max-weight tests to account for the new return shape of _scan_maxwgt_range and ensure range-split behaviour remains consistent.

### DIFF
--- a/MADSPIN_SEQUENTIAL_PLAN.md
+++ b/MADSPIN_SEQUENTIAL_PLAN.md
@@ -894,6 +894,59 @@ error model below is wrong or that implementation is; the combination is
 reachable in the code but should not be used until the weight-identity check
 settles it.
 
+#### The option: one knob, four schemes
+
+`sequential_decay`, `sequential_exact` and `sequential_joint_angles` are replaced
+by a single enumerated option -- the schemes are mutually exclusive alternatives,
+not independent switches, and after variant B was dropped the three booleans no
+longer spanned a clean 2x2:
+
+    set unweighting auto | joint | two_stage | sequential | sequential_global_retry
+
+    mode                     mass stage   angle test              a rejection redraws
+    joint                    --           everything at once      everything
+    two_stage                yes          all angles, one bound   the angles only
+    sequential               yes          per particle            that particle
+    sequential_global_retry  yes          per particle            the virtualities too
+
+`sequential_decay` survives as a deprecated alias (`True` -> sequential,
+`False` -> joint, warning once), so cards written against the earlier revisions
+of this branch keep working. `sequential_spin_order` and `sequential_debug` keep
+their names: an ordering and a check, not modes.
+
+**`sequential_exact` was renamed, not kept.** "Exact" advertised a distinction of
+~0.001 GeV on the top lineshape -- the tabulated factor is good to ~0.5%, and the
+lineshape sensitivity is 0.25 GeV per unit fractional slope error -- inside a pole
+approximation whose own error is of order `Gamma/m` ~ 0.9%, three orders of
+magnitude larger. The name would have pushed users to a 2-3x slower scheme for a
+difference nobody can measure. `sequential_global_retry` says what the mode does
+and leaves the accuracy statement to the documentation, where it can carry the
+numbers.
+
+**`auto`** resolves once per run, from the number of decaying particles counted
+where `to_decay` is built -- not per event, since the modes carry different
+bounds and one that changed event to event would be testing against the wrong
+ones:
+
+- **one decaying particle -> `joint`**, in every spinmode. Every split
+  degenerates there: the per-particle test *is* the joint test (section 3), and
+  the mass/angle one only moves the same factors between two stages. Nothing to
+  win, so the identity machinery is pure cost.
+- **PA/onshell -> `sequential`**. `two_stage` and `sequential_global_retry` split
+  the accept/reject at the up-front mass draw, which those modes do not have;
+  asked for explicitly they log why and fall back to `sequential`.
+- **madspin/full -> `two_stage` for two, `sequential` from three.** One bound
+  over all the angles is tighter than the product of per-particle bounds, while
+  testing each particle as it is drawn lets a rejection skip the decays not yet
+  drawn. The first wins while there is little to skip, the second as the chain
+  lengthens.
+
+That last line makes `auto` non-joint for madspin/full, where `two_stage` is
+faster than the joint test (3.25 production densities and 5.74 decay MEs per
+event against 4.46 and 8.92) and agrees with it at +0.23 sigma over eight
+replicas. An explicit setting is always honoured, including the degenerate
+single-particle case, so any of the four stays available as a cross-check.
+
 #### How to compare these numbers (measurement notes, learned the hard way)
 
 **Wall clocks are only comparable within one campaign.** The per-slot scheme
--- a/MADSPIN_SEQUENTIAL_PLAN.md
+++ b/MADSPIN_SEQUENTIAL_PLAN.md
@@ -998,10 +998,42 @@ the error bar before any comparison is made -- and the two error models
 Any future claim at the few-hundredths-of-a-GeV level needs either many more
 replicas or, better, the deterministic check below.
 
-**The check worth doing anyway.** For one (production, mass set, decay set),
-`w_mass * prod_k (w_k/Z_hat_k)` must equal
-`prod_i n_i * |M_prod|^2_on * wgt_joint` up to floating point -- the same
-deterministic identity that verified the decay-reshuffling jacobian to 1.5e-7.
-It settles exactness for every variant at once, with no error model to argue
-about, and it is the only way to separate a residual `Z_hat` effect from an
-implementation bug in a scheme like variant B.
+**Done: the weight identity holds (`sequential_debug`).** New option, offshell
+only: on every accepted chain, recompute the joint weight with the joint code --
+on copies, for the same production event, the same virtualities and the same
+decays -- and compare with the product of the stage weights.
+
+The weights compared are the ones taken *before* any `Z_hat` division, since
+`Z_hat` cancels between the mass stage and the angle stage. What is tested is
+therefore the decomposition itself, not the table. And what is tested is
+*proportionality*, not equality: the two differ by a constant -- the number of
+helicity states, and the normalisation the density path applies to the decay
+matrix elements relative to `calculate_matrix_element` -- which the bounds
+absorb. A constant ratio is the identity, whatever its value; a scheme sampling
+the wrong distribution has a ratio that varies chain to chain.
+
+Measured over 2000 accepted chains each:
+
+    variant A            spread 1.71e-07   ratio 1108198261
+    sequential per-slot  spread 1.55e-07   ratio 1108198255
+    variant B            spread 1.54e-07   ratio 1108198258
+
+The spread is float32 epsilon (1.19e-7) -- the density matrices are
+`complex64`, so that is the floor of the arithmetic and not physics -- and the
+three schemes agree on the constant to nine significant figures. The threshold
+is `density_tolerance`, for the same reason `density_debug` uses it: two
+evaluation routes through single-precision matrix elements cannot agree better
+than that.
+
+**What this settles, and what it does not.** It settles the *weight algebra* in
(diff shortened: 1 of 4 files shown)


oliviermattelaer:
Thanks — two of the three review findings were worth acting on (2aac8f441), one I pushed back on. Since then the branch has moved a good deal; the current state is at the bottom.

## Cache versioning — valid, done

The cache couples the per-slot bounds and the `Z_k` tables that complete them, so reading one back under a schema it was not written with either crashes inside the accept/reject or silently weights the virtualities with somebody else's fit. It now carries a `format` tag plus a structural check of every field the accept/reject dereferences, and a mismatch is **ignored with a warning** rather than raised on — the scan that produced it is reproducible, so re-measuring is always available. It earned its keep the same day: a24008ac1 bumped the format.

## The polyfit — half right, and that half was a live hole

The performance half does not hold: `_weighted_polyfit2` runs **once per slot at the end of the max-weight scan** — two calls on `p p > t t~` — and there are no shared moments to precompute, each slot's coming from its own bins.

The numerical half was real. The degeneracy test was an absolute floor of `1e-30` on the pivot, while the fit variable `ln(m/pole)` spans ±0.13, so the `x^4` moments start four orders of magnitude below the `x^0` ones and an absolute floor means something different for every resonance. It let singular systems through and returned garbage rather than refusing — every bin at one virtuality gave `[-1.72, 32.0, 0]`, a slope of 32 in `ln(m/pole)`. The tolerance is now relative to the matrix scale; the caller already treats `None` as "no table".

## Extracting the `Z` helpers — declined

144 lines of a 5595-line file, sitting in a family of 23 sibling methods that implement exactly this machinery (`_density_basis`, `_partial_density_contraction`, `_offshell_production`, `_draw_mass_value`, …). Moving one feature's four helpers out while every sibling stays makes the scheme harder to follow, not easier. They touch only `self.banner`, so the extraction stays available as a mechanical move; if the goal is a smaller file the honest unit is the whole density/sequential scheme (~1000 lines), which deserves its own change and its own A/B.

---

# Current state of the branch

Ten commits. Three further findings and an API change since the review.

### The mass weight was missing `|M_prod|^2` on shell (a24008ac1, Olivier)

The mass-set weight had an offshell production ME in the numerator with nothing under it, where the joint weight divides by the on-shell one. Not a bias — that denominator depends on the production event alone, which the chain never redraws, so it cancels between the weight and its bound. But `C_mass` is one number shared by every production event, so without it the absolute scale of `|M_prod|^2` rides inside the weight: the bound is set by the loudest kinematics, the quiet ones pay in acceptance, the loud ones are truncated. `C_mass` 17.1 → 3.09, mass sets per accepted event 26.2 → 3.2, weights above their bound 10 → 1.

### The weight identity, checked deterministically (eda57c8e1)

New `sequential_debug`: on every accepted chain, recompute the joint weight with the joint code — on copies, same production event, same virtualities, same decays — and compare with the product of the stage weights. What is compared is *proportionality*, not equality (the two differ by a constant the bounds absorb), and the weights are taken before the tabulated factor, which cancels between the stages. Over 2000 chains each, the ratio is constant to **1.5–1.7e-07** — float32 epsilon, the density matrices being `complex64` — and the three non-joint schemes agree on the constant to nine significant figures. That settles the weight algebra: no missing jacobian, no wrong normalisation. It does not settle the sampling, which additionally needs nothing self-normalising left uncompensated.

### The tabulated factor is converged at the default probe depth (3a6d52098)

Raising `max_weight_ps_point` 500 → 2500 (187500 samples per slot) moves the table's endpoints by under 1% and reproduces the analytic `(m/M) Gamma(m)/Gamma(M)` to 0.1–0.2%. The lineshape residual then *flipped sign* rather than shrinking (−1.1σ → +0.9σ), so it was noise: pooling eight replicas gives **+0.003 ± 0.011 (+0.23σ)** against joint. This retires an earlier "all replicas sit low" observation of mine, which was a fluke of one set.

### One option for four schemes (a8e22025e, 2cff6bb0a)

`sequential_decay`, `sequential_exact` and `sequential_joint_angles` are replaced by

```
set unweighting auto | joint | two_stage | sequential | sequential_global_retry
```

| mode | mass stage | angle test | a rejection redraws |
|---|---|---|---|
| `joint` | — | everything at once | everything |
| `two_stage` | yes | all angles, one bound | the angles only, so the production reshuffling and its density matrix are reused |
| `sequential` | yes | per particle | that particle |
| `sequential_global_retry` | yes | per particle | the virtualities too |

`sequential_decay` survives as a deprecated alias (`True → sequential`, `False → joint`).

`sequential_exact` was renamed rather than kept: "exact" advertised a ~0.001 GeV distinction on the lineshape inside a pole approximation whose own error is ~`Gamma/m` ≈ 0.9%, and would have pushed users to a 2–3× slower mode for a difference they cannot measure.

**`auto`** resolves once per run, from the number of decaying particles:

| decaying particles | spinmode | mode |
|---|---|---|
| 1 | any | `joint` — every split degenerates, so the identity machinery is pure cost |
| ≥ 2 | PA / onshell | `sequential` — the other two need the up-front mass draw |
| 2 | madspin / full | `two_stage` |
| ≥ 3 | madspin / full | `sequential` |

One bound over all the angles is tighter than the product of per-particle bounds, while a per-particle test lets a rejection skip the decays not yet drawn: the first wins while there is little to skip, the second as the chain lengthens. Since `auto` now depends on the process, every run logs which scheme it used — `MadSpin: unweighting = two_stage (auto, 2 decaying particle(s))`.

### The new default, measured (6c051b6d4)

**This makes `auto` non-joint for madspin/full.** Verified on a card with no `unweighting` line at all, `p p > t t~`, `t > w+ b, w+ > l+ vl`, 10000 events, four replicas each with independent MadSpin seeds, one campaign:

| | replicas (mean of both resonances) | mean |
|---|---|---|
| joint | 173.1852, 173.1866, 173.1769, 173.1784 | 173.1818 ± 0.0024 |
| auto (= `two_stage`) | 173.1474, 173.1787, 173.1938, 173.1616 | 173.1704 ± 0.0101 |

**−0.011 ± 0.010**, i.e. −1.1σ on the replica-scatter error model and −0.7σ on the naive per-run one; pooled with four deep-probe replicas of the same scheme, +0.003 ± 0.011. Decay phase **12.86 s against joint's 14.54 s**, with 3.25 mass sets and 5.74 decay MEs per event against joint's 4.46 trials and 8.92, and one weight over its bound in 10000 events. The `auto` replicas reproduce the explicit `two_stage` ones value for value, which is the check that mattered: `auto` selects a scheme, it is not a fourth path.

### Measurement notes

Two corrections to earlier claims of mine in this thread, both from cross-campaign wall clocks: the per-slot scheme measured 19.74 s in one campaign and 13.33 s in another with **byte-identical counters** — a 48% swing from machine load. Timings are now only quoted within a single campaign with joint as an anchor, and the counters are the quantities to trust. Within one campaign: PA joint 9.17 s, PA sequential 11.19 s, madspin `two_stage` 13.55 s, madspin joint 14.61 s — so full offshell MEs cost ~1.5× the pole approximation.

Separately, and independent of this branch: **PA sequential is 22% slower than PA joint on this process** (each slot trial probes the production reshuffling jacobian) and **logs 11 weight overflows** against the offshell modes' 0–1, so its per-slot bounds are under-estimated there. That is the shipped default path and worth a look on its own.

PR #336, opened 2026-08-15
MadSpin: honour me_frame and beampol in the density spinmodes

oliviermattelaer:
Rebase of `claude/hungry-stonebraker-e6ef3c` onto the landed #334, plus one follow-up on the polarisation convention. Original work by the me_frame/beampol session; I did the merge because their branch was based on a pre-#334 checkpoint and conflicted heavily with the sequential unweighting schemes.

## What it does

`frame_id` (me_frame) and `beampol` were honoured only by the legacy `madspin_v1` path — `decay.py` ships them on stdin to the Fortran driver, which reads them into `/to_me_frame/` and `/to_beampol/`. Every density spinmode (onshell / PA / madspin / full) ignored both, silently.

- **`frame_id` is handled in python.** It is pure kinematics: `frame_id` is the bitmask `sum(2**n)` over the legs whose sum defines the rest frame (the convention `mapid` uncompresses with `btest`), and the momenta are boosted there before the matrix element sees them. `_frame_boost` builds it, `_boost_momenta` applies it inside `get_density`, and the *same* momentum is used for the production density and for every decay contracted against it — otherwise the two sides of the contraction would not share a helicity basis.
- **`beampol` needs the Fortran.** It reweights the initial-state helicity sum, which happens inside the matrix element, so `PY_SET_BEAMPOL(POL1, POL2)` was added to the f2py wrapper and is pushed once per module at initialisation rather than passed on every `get_density` call.

## Two v1 bugs fixed in passing (these change `madspin_v1` output)

- `matrix_standalone_msF_v4.inc` applied the polarisation weight *after* the helicity sum, so the full matrix element stayed unpolarised while the denominator was polarised. No partial beam polarisation ever reached the decay distribution; exactly ±100% happened to work because there the disfavoured helicities get weight zero either way.
- the run_card `polbeam` → `beampol` map used the eva PDF's `[0,1]` left-handed-fraction convention while the Fortran wanted madevent's `/to_polarization/` one.

## One follow-up (ec811a378)

`beampol` was the one place a user met a third convention: the run_card asks for percent (`polbeam1 = 50`), the matrix elements want the internal `sign(1+|polbeam|/100, polbeam)`, and the MadSpin card was asking for that internal value — so `set beampol [1.5, 1.0]` meant a 50% polarised first beam. Since the previous MadSpin convention never worked (see the msF bug above), there is nothing to keep compatible with, so the card now speaks percent like the run_card and the conversion lives in one place, `MadSpinOptions.beampol_me()`, which the density path and all three v1 stdin sites go through.

## Also here

`FourMomentum.boost` returned the boost vector instead of the momentum unchanged when the boost had no three-momentum, where HELAS `boostx` copies. `Event.boost()` would then turn every particle into the boost vector. Reachable whenever the frame-defining system is already at rest — e.g. the `e+ e-` pair of a lepton-collider event — and shared with the joint density path, which boosts the same way. Unit test in `tests/unit_tests/various/test_lhe_parser.py`.

## Merge notes for review

The third commit is **hand-assembled, not a clean replay**, and deserves the closest look. Their branch was based on `a92634d05`, a pre-cherry-pick checkpoint of #334, and also predates madspin_density's `create_f2py_module` rework, so git matched the same changes twice under different hashes and produced misaligned hunks — at one point silently dropping two methods that exist only on the #334 side. I reset the file to the landed version and ported their twelve changes as explicit anchored edits, then verified each is present.

One design point I changed from their description: the frame boost for the offshell branch is computed **once per accepted mass set** inside `_offshell_production`, from the reshuffled production `rho_off` is evaluated at — not per slot. `two_stage` (the new default for two decaying particles) reuses that production density across angle retries, and a per-slot boost would have destroyed exactly the reuse that makes it faster than the joint test.

## Validation

- unpolarised is inert: a 400-event `p p > t t~` smoke reproduces the pre-merge joint result bit-for-bit (`m_top` 173.3373 ± 0.1624), which is the first thing that had to hold since both options default to no-ops;
- all three non-joint unweighting schemes still satisfy the weight identity on the merged tree (`sequential_debug`, ratio constant to 1.7e-07);
- 86 MadSpin unit tests and 298 `various` unit tests pass.

Polarised physics is validated on the original branch (density vs `madspin_v1`); see `UpdateNotes.txt` there.

## Summary by Sourcery

Honour beam polarisation and matrix-element rest-frame options in all MadSpin density modes and fix related polarisation handling bugs.

New Features:
- Add support for beam polarisation and frame_id-based rest-frame boosts in the MadSpin density path, shared consistently between production and decay.
- Expose beam polarisation to the f2py matrix-element libraries and reweight initial-state helicity sums in standalone and split-orders matrix templates.

Bug Fixes:
- Correct polarised-beam reweighting in madspin_v1 so the decay distribution is properly polarised rather than only the denominator.
- Fix FourMomentum.boost so boosting by a momentum at rest leaves the boosted momentum unchanged instead of overwriting it with the boost vector.
- Align beampol convention with the run_card percent-based polbeam settings and madevent’s internal /to_polarization/ representation.

Enhancements:
- Allow MadSpin card values for frame_id and beampol to override production run_card defaults and centralise beampol conversion via MadSpinOptions.beampol_me().
- Cache and reuse frame boosts per production event or accepted offshell mass set to keep production and decay densities in a common helicity basis.
- Extend tests to cover beam-polarisation conventions, frame_id bitmask handling, frame boosts, and boost behaviour for systems already at rest.

Documentation:
- Clarify documentation strings and comments around beam polarisation conventions, frame_id semantics, and their impact on helicity quantisation.

### DIFF
--- a/MadSpin/decay.py
+++ b/MadSpin/decay.py
@@ -2370,7 +2370,7 @@ class decay_all_events(object):
             nb_mc_masses=len(indices_for_mc_masses)
 
             p, p_str=self.curr_event.give_momenta(event_map)
-            stdin_text=' %s %s %s %s %s %s %s \n' % ('2', self.options['BW_cut'], self.Ecollider, decay_me['max_weight'], self.options['frame_id'], self.options['beampol'][0], self.options['beampol'][1]) 
+            stdin_text=' %s %s %s %s %s %s %s \n' % ('2', self.options['BW_cut'], self.Ecollider, decay_me['max_weight'], self.options['frame_id'], self.options.beampol_me()[0], self.options.beampol_me()[1]) 
             stdin_text+=p_str
             # here I also need to specify the Monte Carlo Masses
             stdin_text+=" %s \n" % nb_mc_masses
@@ -2503,7 +2503,7 @@ class decay_all_events(object):
         except KeyError:
             frameid = 6
         try:
-            beampol = self.options['beampol']
+            beampol = self.options.beampol_me()
         except KeyError:
             beampol = (1.0,1.0)
 
@@ -3447,7 +3447,7 @@ class decay_all_events(object):
         """return the max. weight associated with me decay['path']"""
 
         p, p_str=self.curr_event.give_momenta(event_map)
-        std_in=" %s  %s %s %s %s %s %s \n" % ("1",BWcut, self.Ecollider, nbpoints, self.options['frame_id'], self.options['beampol'][0], self.options['beampol'][1])
+        std_in=" %s  %s %s %s %s %s %s \n" % ("1",BWcut, self.Ecollider, nbpoints, self.options['frame_id'], self.options.beampol_me()[0], self.options.beampol_me()[1])
         std_in+=p_str
         max_weight = self.loadfortran('maxweight',
                                path, std_in)
--- a/MadSpin/decay.py
+++ b/MadSpin/decay.py
@@ -2505,7 +2505,7 @@ class decay_all_events(object):
         try:
             beampol = self.options['beampol']
         except KeyError:
-            beampol = (0.5,0.5)
+            beampol = (1.0,1.0)
 
         stdin_text=' %s %s %s %s %s %s %s\n' % ('2', self.options['BW_cut'], self.Ecollider, 1.0, frameid, beampol[0], beampol[1])
         stdin_text+=p_str
--- a/MadSpin/interface_madspin.py
+++ b/MadSpin/interface_madspin.py
@@ -114,6 +114,16 @@ class MadSpinOptions(banner.ConfigFile):
             random.seed(self['seed'])  
             random.mg_seedset = self['seed']  
 
+    def post_set_beampol(self, value, change_userdefine, raiseerror, *opts):
+        """Two values or none: one number is ambiguous (which beam?) and would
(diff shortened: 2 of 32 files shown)


oliviermattelaer:
Review addressed in 576047d5b.

**`beampol_me` on malformed input — valid, and reachable.** `set beampol 50` is accepted by `ConfigFile`, stored as `[50.0]`, and then raises `IndexError: list index out of range` inside `beampol_me`, once the run has reached a matrix element and long after the card was read. Fixed, but at *set* time rather than by normalising: one number is genuinely ambiguous — first beam, or both? — so padding it with an unpolarised second beam would turn a typo into a silently different physics setup. It is now refused when the card is read, with a message that says what to write. An empty list stays legal and means unpolarised. Regression test added.

**UpdateNotes typo** — fixed.

**Duplicated polarisation reweighting in the Fortran templates** — not doing this here. The duplication mirrors madevent's own per-variant template structure (`matrix_standalone_msP/msF`, the split-order variants, the loop templates), and every one of those files is a code-generation template with IOTest reference output pinned to it. Factoring them into a shared helper is a code-generation refactor with a large blast radius, and it should not ride along in a PR whose subject is MadSpin option handling. Worth its own issue if the divergence risk is judged real.

PR #337, opened 2026-08-15
MadSpin: honour the '@' decay grouping tags in the density modes

oliviermattelaer:
Work of the "Verify positional decay assignment, scope @-tag support" session. I am opening the PR because the branch was finished but never pushed; **I have not reviewed or re-run it**, and the description below is taken from its own commit messages.

## What it does

The `@` suffix sorts decay lines into groups meant to be used together — the semi-leptonic ttbar idiom, where only the two charge assignments exist and no fully leptonic or fully hadronic event is produced. Only `madspin_v1` implemented it, by generating one decayed matrix element per group. The density modes fill one decay pool per particle and per channel and drew each channel independently, so the tag had nowhere to attach — and it was swallowed **silently**, because the decay-ME generation appends an `@` process number of its own and MG5 absorbs the user's as the sub-decay's. The card did not mean what it looked like it meant, with no warning.

The branch fixes that in three steps: warn that the tags are ignored, then parse and validate them into a layout the rest of the code can act on, then honour them in PA, onshell and madspin/full.

## Validation it reports

- a grouped run against the merge of one run per group, 20000 production events decayed off the same production file, reference built pairwise so the production kinematics are identical on both sides;
- the `n_part = 2` case (`p p > t t~ t t~`), where a group supplies two lines per pdg and the positional rule deals them out inside the group;
- PA: 142.54946 pb grouped against 71.26739 + 71.24537 = 142.51276 dedicated, ratio 1.000258; splitting the grouped sample by group and comparing each half with its own dedicated run, three of thirty comparisons land at 2.0–2.7 sigma and none reproduces at another seed;
- grouping forces the joint accept/reject, and that fallback is a no-op: a grouped card at the sequential default produces event records byte-identical to the same card with the joint test.

It also carries `doc/madspin_decay_groups.md` (544 lines) and 346 lines of unit tests.

## Needs work before merging

**The branch is based on 275462e52, i.e. madspin_density before #334 landed**, and conflicts in all three shared files (`MadSpin/decay.py`, `MadSpin/interface_madspin.py`, `tests/unit_tests/madspin/test_madspin.py`). #334 rewrote large parts of `interface_madspin.py` — 954 insertions in that file alone — so this needs a real merge, not a fast-forward.

One concrete consequence: this branch talks to the **old option API**. #334 replaced `sequential_decay` / `sequential_exact` / `sequential_joint_angles` with a single `unweighting = auto | joint | two_stage | sequential | sequential_global_retry`, so the "grouping forces the joint accept/reject" logic and its tests need re-expressing against `unweighting`.

## An independent confirmation of #334, worth recording

While validating the grouping, this session found — and its final commit documents — that `sequential_decay True` biased the top lineshape under madspin: one ungrouped card, the same production events, only the accept/reject changed, giving mean `m(top)` 173.16870 joint against 172.94647 sequential, 7.0 sigma with KS p = 0.0000, while every angular observable stayed within 1.0 sigma. PA showed the same at 2.1 sigma.

That is the same `Z_k` bias #334 was opened to fix, found independently and by a different route (a grouping study rather than a lineshape A/B). #334 is merged, so this branch will inherit the fix through the merge above.

## Summary by Sourcery

Support and validate MadSpin `@` decay grouping tags in density spin modes, ensuring grouped decays are correctly applied, reported, and fall back safely when unsupported.

New Features:
- Enable per-event decay grouping based on `@` tags in density-based spin modes (PA, onshell, madspin/full), drawing one group per event and assigning its channels to all relevant particles.
- Introduce support for grouped decays with multiple identical parents, using positional assignment within each group while keeping branching ratios consistent.

Bug Fixes:
- Prevent user `@` grouping tags from being silently misinterpreted as MG5 process numbers by stripping them before matrix-element and decay-command generation.
- Warn when `@` decay grouping tags are present but cannot be honoured in the current spin mode or configuration, avoiding misleading cards and unintended ungrouped behaviour.
- Clamp partial widths that exceed the total width to protect branching-ratio calculations and avoid hidden distortions, including for multi-channel positional cases.

Enhancements:
- Add infrastructure to parse, validate, and resolve decay-group layouts from MadSpin cards, including rectangular-shape checks against production events and multiparticle safeguards.
- Extend branching-ratio computation to handle grouped decays via per-group rate factors and probabilities, integrating group rates into the global BR factor.
- Force joint accept/reject when decay groups are present to preserve correct group fractions, and guard sequential accept/reject from being used with grouped decays.
- Refine decay-pool generation to track per-channel widths and support group-aware pool sizing and assignment multiplicity for identical channels.
- Update decay drawing logic to support group-aware channel selection while preserving existing positional and cross-section-weighted rules for ungrouped decays.

Documentation:
- Add `doc/madspin_decay_groups.md` describing the semantics of `@` decay grouping tags, supported card shapes, algorithmic design, limitations with sequential/two-stage unweighting, and validation results.

Tests:
- Add unit tests covering `@` tag parsing and warning behaviour across spin modes, decay-group layout and validation against production multiplicities, group drawing and per-particle channel selection, assignment multiplicity factors, and grouped branching-ratio and probability computations.
- Extend existing MadSpin tests with stubs and helpers for group drawing and decay selection to validate grouped behaviour with single and multiple identical parents.

### DIFF
--- a/MadSpin/decay.py
+++ b/MadSpin/decay.py
@@ -4440,6 +4440,8 @@ class decay_all_events_onshell(decay_all_events):
         decay_text = []
         for decays in self.mscmd.list_branches.values():
             for decay in  decays:
+                # MadSpin's own '@' grouping tag, not something MG5 should see
+                decay = self.mscmd._split_group_tag(decay)[0]
                 if '=' not in decay:
                     decay += ' QCD=99'
                 if ',' in decay:
@@ -4459,6 +4461,8 @@ class decay_all_events_onshell(decay_all_events):
         decay_text = []
         for decays in self.mscmd.list_branches.values():
             for decay in  decays:
+                # MadSpin's own '@' grouping tag, not something MG5 should see
+                decay = self.mscmd._split_group_tag(decay)[0]
                 if '=' not in decay:
                     decay += ' QCD=99'
                 if ',' in decay:
@@ -4467,7 +4471,7 @@ class decay_all_events_onshell(decay_all_events):
                     decay_text.append(decay)
         decay_text = ', '.join(decay_text)
 #        commandline = ''
-        
+
         for proc in processes:
             if not proc.strip().startswith(('add','generate')):
                 proc = 'add process %s' % proc
@@ -4482,10 +4486,18 @@ class decay_all_events_onshell(decay_all_events):
         i=0
         for processes in self.list_branches.values():
             for proc in processes:
+                # Drop MadSpin's own '@' grouping tag first: this line appends a
+                # process number of its own, MG5 binds that at the top level and
+                # would absorb the user's as the process number of the
+                # *sub-decay*. Nothing would fail -- the amplitude is the same --
+                # but the tag is MadSpin bookkeeping and has no business
+                # reaching MG5. (madspin_v1 strips it the same way, decay.py
+                # get_all_ME step 6.)
+                proc = self.mscmd._split_group_tag(proc)[0]
                 newproc = "add process %s @%i --no_warning=duplicate --standalone;" % (proc,i)
-                commandline += self.adapt_decay(newproc) 
+                commandline += self.adapt_decay(newproc)
                 #commandline+="add process %s @%i --no_warning=duplicate --standalone;" % (proc,i)
-                i+=1 
+                i+=1
         return commandline
 
     def compile(self):
--- a/MadSpin/interface_madspin.py
+++ b/MadSpin/interface_madspin.py
@@ -4363,7 +4363,7 @@ class MadSpinInterface(extended_cmd.Cmd):
         logger.info("*****************************")
         logger.info("Probing the first %s events with %s phase space points"
                     % (nevents, nb_ps_point))
-        # sequential_decay never reaches here with fixed_order (it falls back to
+        # a non-joint unweighting never reaches here with fixed_order (it falls back to
         # the joint accept/reject), so the events are plain, not event-groups.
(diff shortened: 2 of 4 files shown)


oliviermattelaer:
Review addressed in 51289b39b.

**`_clamped_partial_width`** — the docstring was wrong, the code is right, so I fixed the docstring rather than the logic. The two regimes are not new policy: they are exactly what `run_onshell` has always done (`if pwidth > 1.01*totwidth: warn; elif pwidth > totwidth: cap`), which this helper factors out unchanged. They are also deliberate — up to 1% over, the excess is Monte Carlo noise in the width measurement, so capping is free and keeps the branching ratio at most 1; beyond that the param_card total genuinely disagrees with what was generated, and capping would quietly reshape the normalisation to match a card that is wrong while hiding the disagreement. Always clamping, as suggested, would be a silent change to branching-ratio normalisation shipped inside a PR about decay grouping.

**Targeted tests for it** — added, one per regime, pinning the behaviour so the next reader does not have to re-derive it from `run_onshell`.

**Centralising the `@`-tag stripping** — already the case: every MadSpin group-tag site goes through `_split_group_tag` (six call sites in `interface_madspin.py`, one in `decay.py`). The remaining `split("@")` calls in `decay.py` are pre-existing code handling MG5's *own* process numbers, which is a different thing and unrelated to the grouping tags.

Note for reviewers: this branch is still based on the pre-#334 `madspin_density` and needs a merge before it can land — see the PR description.

PR #346, opened 2026-08-18
MadSpin: factor the spinmode branching out of the unweighting decisions

oliviermattelaer:
Follow-up to the review remark that the `_sequential_upfront` / `_unweighting_mode` interaction had become hard to follow. Strictly behaviour-preserving: the spinmode-specific branching is pulled out into named predicates so the up-front/offshell decisions can be read rather than re-derived.

### New helpers

All private on `MadSpinInterface`; no public or option-facing name changed, no log message changed.

| helper | contract |
|---|---|
| `_density_pole_approximation()` | `spinmode in ['PA','onshell']` — density matrix at onshell momenta |
| `_density_do_reshuffle()` | `spinmode == 'PA'` — a PA run that still reshuffles onto sampled virtualities |
| `_density_needs_reshuffle(in_density_mode)` | `in_density_mode and (not PA or do_reshuffle)`; the argument is the caller's own "am I in density mode" (`density_method` before generation, `self.generate_all.mode == 'density'` after) |
| `_spinmode_has_density()` | `spinmode in ['PA','onshell','madspin','full']`, expressed as PA-family union offshell-family |
| `_is_upfront_scheme(mode)` | static; `mode not in ('joint','sequential_with_mass')` — does the scheme have a mass-set stage |
| `_auto_unweighting_mode()` | what `unweighting='auto'` resolves to before the fallbacks: `sequential` under PA/onshell; offshell `joint` at <=2 decaying particles, `sequential` from 3 |

The `auto` benchmark docstring (the `p p > w+ j` / `t t~` / `t t~ z` / `t t~ t t~` numbers, the `Tr(rho_off)/|M_prod|^2` argument, the `two_stage` note) was moved **verbatim** onto `_auto_unweighting_mode`, with a pointer left behind. `_sequential_upfront` keeps its spinmode paragraph; only the "which schemes" sentence moved to `_is_upfront_scheme`.

Also: `density_do_reshuffle` in `decay_events` was computed and never used — it only ever fed `density_needs_reshuffle`, which now derives it internally — so it is gone.

### Evidence that behaviour is unchanged

**Exhaustive before/after snapshot.** A throwaway script borrows the real unbound methods onto a stub (via `inspect.getattr_static`, so staticmethods stay static) and enumerates 9 spinmodes (all 7 declared + `bridge` + an unknown name) x 6 `unweighting` values x 6 decay multiplicities (0,1,2,3,4,7) x `fixed_order` x decay-groups-present x `density_method` = **2610 rows**. Each row records `_unweighting_mode`, `_sequential_active`, `_sequential_upfront`, `_sequential_offshell`, the default-argument variants, and `_sequential_pool_ladder` over 7 `to_decay`/`nb_event` shapes; plus the `density_*` triple per spinmode x `density_method`.

```
$ diff before.json after.json && echo "SNAPSHOT IDENTICAL (2610/2610)"
SNAPSHOT IDENTICAL (2610/2610)
```

Not vacuous: 0 error rows, 114 non-empty ladders, all five schemes reachable.

**Parametrized unit test.** `TestUnweightingDecisionTable` (10 tests) checks `_unweighting_mode` over the full 1296-case product against a `_reference_mode` written from the `unweighting` option comment rather than read off the implementation, plus the `auto` table, the with-mass fallback, the family predicates and the pool ladder.

**Runs.** No pytest in this environment; the repo convention is `tests/test_manager.py`.

```
$ python3 tests/test_manager.py -p tests/unit_tests/madspin
Ran 141 tests in 10.316s
OK
```

(131 before this change, so the 10 new tests are additive and nothing regressed.)

Full unit suite: `Ran 918 tests ... FAILED (errors=1)` — `test_DensityMatrixObservables22`, `ModuleNotFoundError: No module named 'scipy'` in `madgraph/various/Density_functions.py:1035`. Pre-existing environment gap, untouched by this diff.

### Notes for review

- **Four existing test stubs had to be touched.** They borrow a hand-listed set of unbound methods, so the new helper calls inside `_unweighting_mode` broke 26 tests with `AttributeError`. Rather than repeat six lines per stub there is now a module-level `_borrow_decision_helpers(locals())` (`test_madspin.py:52`) called from each. Any new stub that borrows `_unweighting_mode` needs that call too.
- **Not verified by execution:** `tests/parallel_tests/test_madspin_factory.py` asserts the resolved scheme by parsing the `MadSpin: unweighting = ...` log line, and needs real MG5 runs. `_announce_mode`, `_log_once` and every message string are byte-identical so the announced line cannot have moved — but that is reasoning, not a run.
- `draw_mass = (spinmode == 'PA' and nb_prod_final > 1)` was deliberately left alone: same predicate, but it reads as "does this mode draw a mass" rather than "does it reshuffle", and folding it in would have obscured rather than clarified.

## Summary by Sourcery

Clarify MadSpin’s spinmode-dependent unweighting decisions by factoring them into reusable predicates while preserving existing behavior.

Enhancements:
- Extract spinmode, density, automatic unweighting, and upfront-scheme decisions into named MadSpin predicates without changing behavior or user-facing messages.

Tests:
- Add exhaustive parametrized coverage for unweighting decisions, spinmode predicates, fallbacks, and sequential pool ladders across supported combinations.

### DIFF
--- a/MadSpin/interface_madspin.py
+++ b/MadSpin/interface_madspin.py
@@ -2430,13 +2430,8 @@ class MadSpinInterface(extended_cmd.Cmd):
         self.error *= self.branching_ratio
         
 
-        density_pole_approximation = self.options['spinmode'] in ['PA', 'onshell']
-        density_do_reshuffle = self.options['spinmode'] == 'PA'
-        density_needs_reshuffle = (
-            density_method
-            and (not density_pole_approximation
-                 or density_do_reshuffle)
-        )
+        density_pole_approximation = self._density_pole_approximation()
+        density_needs_reshuffle = self._density_needs_reshuffle(density_method)
 
         # 3. generate the various matrix-elements
         time_me_generation = time.time()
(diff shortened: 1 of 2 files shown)


oliviermattelaer:
Merged the current `madspin_density` tip (through #347) into this branch and resolved the two conflicts that #344 created:

- `_unweighting_mode` docstring: #344 rewrote the `two_stage` paragraph while this PR *moves* that docstring onto the new `_auto_unweighting_mode`. Kept the relocation and carried #344's new wording (the `hidden_unweighting_modes` note) into the moved text, so nothing #344 said is lost.
- `tests/unit_tests/madspin/test_madspin.py`: add/add at the end of the file, where this PR and #345 each appended a test class. Kept both.

No functional change on either side of the resolution. `./tests/test_manager.py test_madspin -t0` -> `Ran 147 tests ... OK`.

PR #349, opened 2026-08-18
MadSpin: restrict the density convolution to the production polarisation

oliviermattelaer:
Restricts the production/decay density-matrix convolution according to the polarisation braces on the decaying particle in the **production** process.

Until now all density spinmodes contracted the full double sum `W = sum_i sum_j rho_prod(i,j) rho_dec(i,j)`. Now, per decaying particle:

- `{0}`, `{+}`/`{R}`, `{-}`/`{L}` — a single helicity X — keep only `rho_prod(X,X) rho_dec(X,X)`;
- `{T}` keeps the double sum over the `-1/+1` block and drops the `0` row and column;
- no brace: unchanged, bit-for-bit.

The rule is uniform — entry (i,j) of particle *k* survives iff **both** i and j are allowed for *k* — which produces the diagonal-only and transverse-block cases automatically and composes per index for several decaying particles.

## The restriction genuinely changes results

The key question was whether MadSpin's production ME already carries the polarisation projection. **It does not.** Verified at Fortran level: for `u u~ > w+{0} w-` (NCOMB=12, every NHEL row has NHEL(3)=0) the density matrix comes back **bit-identical** to the unpolarised `u u~ > w+ w-`, including the `rho(-,-)` and `rho(+,+)` entries `{0}` forbids, because `GET_ALL_INTER` overrides the polarisation at amplitude level.

Confirmed again end to end: under the **old** code a `t{R}` run was byte-identical to the unpolarised run — the brace was ignored outright.

## Two pre-existing bugs fixed

1. **A polarised production hung MadSpin indefinitely.** `GET_DENSITY` picks its NHEL rows by matching against the *first* `ALLOW_HEL` combination, and with `hel_dict = {2:[1,-1], 3:[-1,0,1]}` a `{0}`, `{L}` or `{-}` production has no matching row, so the whole production density matrix was identically zero. A zero `rho_prod` rejects *every* accept/reject trial, so MadSpin does not error — it loops forever regenerating decay pools:

   ```
   old code, p p > t{L} t~ :   killed at 600 s, 131 decay-pool regenerations, no output
   old code, p p > w+{0} w- :  exit=124 (timeout 240 s), 41 regenerations, NO DECAYED FILE
   new code, same inputs/seed: exit=0, decayed file written, seconds
   ```

   Only `{R}`/`{+}` on a fermion happened to work. Fixed by having `_density_basis` reorder each polarised particle's helicity basis so an allowed helicity leads.

2. **`adapt_production` matched the whole token**, so a polarised leg `t{L}` never received its off-shell star. Now the name before the brace is matched; the run log shows MG5 receiving `generate p p > t{L}* t~*` and producing a genuinely polarised ME (`P1_gg_tLtx` has NCOMB=8 vs 16 unpolarised, `NHEL(3)` pinned to `-1`).

## End-to-end validation

Event-block hashes (whole-file hashes differ only by banner paths/dates):

```
old_nopol    38e1065c...  new_nopol  38e1065c...   -> IDENTICAL (no-brace unchanged)
old_tR       38e1065c...  new_tR     7507ca05...   -> DIFFER (old ignored the brace)
```

Physics against the analytic distributions (theta in the parent rest frame vs the parent's lab direction — the helicity axis):

```
                e+ from t                        e- from t~ (no brace: control)
old_nopol    <cos> = +0.015 +- 0.061           <cos> = +0.011 +- 0.058
new_nopol    <cos> = +0.015 +- 0.061           <cos> = +0.011 +- 0.058
old_tR       <cos> = +0.015 +- 0.061           <cos> = +0.011 +- 0.058
new_tR       <cos> = +0.409 +- 0.045  [+1/3]   <cos> = -0.060 +- 0.055
new_tL       <cos> = -0.352 +- 0.051  [-1/3]   <cos> = +0.072 +- 0.059
```

```
W+ -> e+ ve, 2000 generated events each   <cos^2>          expected
p p > w+ w-      (madspin)                0.347 +- 0.007   SM mixture
p p > w+{T} w-   (madspin)                0.393 +- 0.007   2/5 = 0.400
p p > w+{T} w-   (onshell)                0.398 +- 0.007   2/5
p p > w+{0} w-   (madspin)                0.198 +- 0.005   1/5 = 0.200
p p > w+{0} w-   (PA)                     0.208 +- 0.005   1/5
```

`{T}` lands within 1 sigma of 2/5, `{0}` within 0.4 sigma of 1/5, and `{0}` gives `<cos> = -0.001 +- 0.010` — forward/backward symmetric, as sin^2(theta) requires. The `t~` control stays compatible with zero in every polarised run, confirming the mask is per particle. Three density spinmodes exercised: `madspin`, `PA`, `onshell`. The `t{L}` run also covers the off-shell polarised amplitude enabled by fix 2 — it behaves, not merely parses.

## Decay-side braces now error clearly

```
InvalidCmd: MadSpin: polarization (the {...} braces) is not supported on a 'decay'
line with spinmode=madspin. Only the polarization of the production process is
taken into account by the density spin modes (madspin/full/PA/onshell); use
spinmode=none or spinmode=madspin_v1 for decay-side polarization.
```

## Implementation

`MadSpin/decay.py` — the restriction rides on the `DensityMatrix` rather than on the call, so `scalar_multiplication` and `trace()` pick it up at every contraction site with no call-site change, including `_partial_density_contraction` in the sequential accept/reject. Masks are cached per `(basis_id, restriction)`, so they are built once per basis and never per event, and both existing fast paths are preserved.

`MadSpin/interface_madspin.py` — `_production_polarization()` parses the banner proc_card through `mg5cmd.extract_process`, so brace semantics cannot drift from MG5's; `_density_basis` gains a `hel_restriction` key; the three production `get_density` call sites pass it. `do_launch` validates the braces up front rather than on the first event inside a worker.

**Normalisation:** `prod_diag` is now the *restricted* trace. Required, not cosmetic: with the identity decay density the weight still averages to `1/n`, so `N_0 = Tr(rho)/prod n_i` and the accept/reject bookkeeping is untouched. Decay diagonals stay unrestricted — decay events come from the full decay ME.

`Ran 148 tests ... OK` (131 before; 17 new).

## Two things flagged, deliberately not fixed here

- **A vanishing production density matrix makes MadSpin hang rather than fail.** After this PR no polarisation input reaches that state, but any other cause would. A guard that bails out with a clear error when `Tr(rho_prod)` is zero across a whole pool regeneration is worth a follow-up; it touches the shared unweighting loop and does not belong in this diff.
- On the *first* old-code `t{L}` attempt, the runaway run also deleted sibling files in its parent directory. Not reproduced in a sandboxed layout, and never seen on any healthy run — believed to be collateral from the runaway pool regeneration rather than an independent bug, but recorded rather than asserted harmless. It only occurs on the path this PR fixes.

## Not verified

Same-pdg-different-polarisation (`p p > w+{0} w+{T}`) is refused with an explicit error rather than supported: the density basis identifies particles by pdg-and-production-order, which cannot be reliably matched to process-line order.

## Summary by Sourcery

Apply production polarisation projections consistently throughout MadSpin density calculations and expose optional per-polarisation event weights.

New Features:
- Restrict production/decay density-matrix convolutions according to per-particle production polarisation braces, including longitudinal, transverse, and multi-particle assignments.
- Add optional per-polarisation LHEF weights for vector and fermion decaying particles, with cartesian combinations, production-brace intersections, banner declarations, and deprecated-option compatibility.

Bug Fixes:
- Fix polarised production density matrices that previously became zero and caused MadSpin to loop indefinitely.
- Fix polarised production process adaptation so off-shell stars are applied to legs whose names include braces.
- Honor the production matrix-element frame for polarised density contractions and correct frame-metadata unpacking.
- Reject unsupported decay-side polarisation braces clearly in density spin modes.

Enhancements:
- Carry helicity restrictions through density-matrix contractions, tensor products, traces, and sequential accept/reject normalization with cached masks while preserving unrestricted behavior.
- Validate and parse production polarisation consistently with MG5, including same-particle assignments and incompatible subprocess patterns.

Documentation:
- Document production-polarisation restrictions, frame handling, normalization, and validation behavior in the sequential-density plan.

Tests:
- Add extensive unit coverage for density restrictions, production-polarisation plumbing, same-PDG assignments, frame handling, and optional polarisation-weight generation.

### DIFF
--- a/MADSPIN_SEQUENTIAL_PLAN.md
+++ b/MADSPIN_SEQUENTIAL_PLAN.md
@@ -63,19 +63,40 @@ Two things this is NOT:
   they build, so `rho_prod` comes back **fully unpolarised** in those indices
   even for a polarised process -- verified by evaluating `M0_GET_DENSITY`
   directly for `p p > t{L} t~` and for `p p > t t~`: identical entries,
-  including `rho(+1,+1)`. The restriction therefore changes results.
+  including `rho(+1,+1)`. End to end, `p p > t{R} t~` before this change gave
+  event blocks *byte-identical* to `p p > t t~`: the brace was ignored outright.
 - it is not free of a basis reordering. `GET_DENSITY` selects the rows of the
   process' `NHEL` table by matching them against the *first* `ALLOW_HEL`
   combination, and a polarised process has no `NHEL` row outside its
   polarisation. With the default `hel_dict` order (`[1,-1]`, `[-1,0,1]`) a
   `{L}`/`{-}`/`{0}` production matched nothing and handed back an identically
-  zero density matrix. `_apply_production_polarization` therefore puts an
+  zero density matrix -- and a zero `rho_prod` rejects *every* accept/reject
+  trial, so MadSpin did not fail, it **looped forever** regenerating decay-event
+  pools (observed: `p p > t{L} t~` and `p p > w+{0} w-` both spin indefinitely
+  on the pre-change code). `_apply_production_polarization` therefore puts an
   allowed helicity first in that particle's basis; the order is untouched when
   there is no brace.
 
 Polarisation on a **decay** line is rejected outright in the density spin modes
 (`do_decay`): the braces there would restrict the decay matrix element that
 defines the branching ratio, not the density matrix that is contracted.
+
+Validated end to end against the analytic decay distributions (`spinmode` in
+parentheses; theta measured in the parent rest frame against the parent's lab
+direction, which is the axis the helicity is quantised along):
+
+| production | observable | measured | expected |
+|---|---|---|---|
+| `p p > t t~` | `<cos>` of e+ from t | +0.015 +- 0.061 | ~0 (unpolarised) |
+| `p p > t{R} t~` (madspin) | idem | +0.409 +- 0.045 | +1/3 |
+| `p p > t{L} t~` (madspin) | idem | -0.352 +- 0.051 | -1/3 |
+| `p p > w+ w-` (madspin) | `<cos^2>` of e+ from W+ | 0.347 +- 0.007 | SM mixture |
+| `p p > w+{T} w-` (madspin / onshell) | idem | 0.393 / 0.398 +- 0.007 | 2/5 |
+| `p p > w+{0} w-` (madspin / PA) | idem | 0.198 / 0.208 +- 0.005 | 1/5 |
+
+In every polarised run the *un*polarised partner in the same event (the `t~`,
+which carries no brace) stayed compatible with zero, confirming the mask is
+per particle. A no-brace run is byte-identical to the pre-change code.
 
 ### The partial weight
 
--- a/MADSPIN_SEQUENTIAL_PLAN.md
+++ b/MADSPIN_SEQUENTIAL_PLAN.md
@@ -82,8 +82,9 @@ Polarisation on a **decay** line is rejected outright in the density spin modes
 defines the branching ratio, not the density matrix that is contracted.
 
 Validated end to end against the analytic decay distributions (`spinmode` in
-parentheses; theta measured in the parent rest frame against the parent's lab
-direction, which is the axis the helicity is quantised along):
+parentheses; theta measured in the parent rest frame against the parent's
+direction **in the `me_frame` frame**, which is the axis the helicity is
+quantised along -- see the next section, which is what fixed that axis):
 
 | production | observable | measured | expected |
 |---|---|---|---|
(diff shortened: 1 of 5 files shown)


oliviermattelaer:
⚠️ **Do not merge this PR on its own — #355 is a prerequisite.**

A follow-up investigation (#355, stacked on this branch) found that the polarisation here is defined on the **lab** axis, while MG5 defines it on `run_card: me_frame` (default `[1,2]` -> `frame_id=6` -> the partonic CM). A polarised matrix element is not Lorentz invariant, so the frame decides *which helicity* `{0}` names.

This was harmless before: `_frame_boost` disabled itself unless the beams were polarised, which was correct because the full contraction is a trace and a boost is a unitary basis change that cancels between rho_prod and rho_dec. **This PR breaks that argument** — `set_hel_restriction` is a projection, and projections do not commute with basis changes.

Measured on `p p > w+ w-`, 10000 events: read as a decomposition on MG5's axis, this branch alone gives f0 = 0.63 for a nominally 100% longitudinal sample and 0.11 for a nominally 0% one — about a third of the requested purity lost to the frame.

#355 also fixes a second, independent bug it uncovered: `_frame_boost` unpacked four values from `get_pdir`, which has returned five since `6ba177c56`, so the entire `me_frame` path in the density modes had never executed and `polbeam1`/`polbeam2` crashed in a density spinmode. The unit tests were green only because the test stub returned the old arity.

PR #351, opened 2026-08-18
MadSpin: pure-interference mode

oliviermattelaer:
**Stacked on #349.** Was an assessment plus an inert primitive; now the mode is **implemented and validated end to end**.

`set pure_interference w+ = 0 T` writes a sample whose total cross-section is zero by construction, whose event weights keep the unpolarised magnitude, and whose signs follow the sign of the interference convolution.

## Design, from the assessment in `MADSPIN_SEQUENTIAL_PLAN.md` §13

- **The contraction is real without taking a real part.** A per-particle entry may be a `(P, D)` pair; `_restriction_row_mask` builds `(bra in P & ket in D) | (bra in D & ket in P)`. The union is inside one particle's factor, so each factor is transposition-closed and so is the product. Taking `Re` of a half-sum would be off by a factor 2.
- **Zero cross-section is a theorem**: the interference block has no diagonal entry, so the restricted trace is exactly zero.
- **Contraction and normalisation are therefore decoupled.** `DensityMatrix.hel_restriction_trace` is consulted by `trace()`/`normalized()` **only** for a cross restriction — `_trace_restriction()` returns a symmetric restriction untouched, so no pre-existing path can move and no sentinel is needed.
- **Accept/reject reworked for signed weights.** Redraw-until-accept is *statistically wrong* here: it is correct today only because `<wgt>` is a constant per production event, but that mean is zero and what varies is `<|wgt|>` — exactly the local interference size. So: one draw, accept on `|w|/maxwgt`, `continue` on rejection. `_joint_maxwgt_range` bounds `|w|`. Both gated on the mode. The sign rides onto `full_evt.wgt` and every `parse_reweight()` entry.
- **Sequential is forced to `joint`**: every partial contraction against `DensityMatrix.identity` is zero, so all prefixes weigh zero.
- **Syntax**: a plain string param, `;`-separated for several particles. Not a dict param — `banner.ConfigFile`'s dict path does `x, y = value.split()` and chokes on a four-token value.

## Measured end to end

`p p > w+ w-`, `spinmode=madspin`, leptonic decays, `pure_interference w+ = 0 T`, `nb_core > 1`.

**Zero cross-section**, five seeds, 100k production events combined:

```
S = -28.75    sqrt(sum w^2) = 61.98    z = -0.46
per-seed z:  +2.69, -0.75, -2.04, -0.00, -0.83      mean -0.18 +- 0.79
zero overweight trials anywhere
```

The first run gave `+2.69` and was not taken at face value — the four extra seeds are what establish it as a fluctuation rather than a bias, and are the reason the 5-sigma threshold is the right one.

**Both signs present**: 629+/537- (seed 42), 620+/694- (seed 99).

**`|w|` matches the unpolarised magnitude**: exactly `0.79865722 = sigma * BR`, relative difference `0.000e+00` against the unpolarised run at the same seed. Across seeds it moves by `4e-5`, the per-run MC estimate of the BR. The interference is carried entirely by the ~6% keep rate.

**No-op when unset**: same card against the pre-implementation tree and after → identical 135,011,930-byte file, banner included.

**Tests**: `Ran 184 tests ... OK` (baseline on this branch after the base merge re-measured as 161, not the ~159 I had been told).

Also added: the overweight-trial counter from §13.8, which is what makes a non-zero `z` diagnosable in the first place.

## Two reconciliation items for the maintainer

**1. `_frame_boost` guard.** The cross restriction is a projection, so it needs MG5's `me_frame` axis just as #355's braces do — but a pure-interference run has an *unpolarised* production, so #355's condition would switch the boost off. Clause added:

```python
if (self._beampol() is None and not self._production_polarization()
        and not self._pure_interference()):
    return None
```

A `NOTE (reconciliation)` comment sits at that line, and plan §13.10 step 9 records that the clause belongs inside `_needs_frame_axis()` once the parallel branch (#357) merges.

**2. PR #354's `_dead_trial` guard will discard the negative half.** #354 is **not in this base**, so there was no interaction to test and none was guessed at. What to expect on merge: that guard classifies zero/negative/NaN weights as dead trials, but here a negative weight is the *normal, physically meaningful* case — roughly half of all accepted events. Left as-is it would discard them and destroy the cancellation that makes the sample sum to zero. It needs the same mode gate the accept test now carries: test `abs(wgt)`, treat only zero/NaN as dead. Its zero test may also need care, since a genuinely zero interference weight is common near the sign flip.

## Boundaries, recorded in plan §13.9/§13.12

- Only `spinmode=madspin` was run end to end; `PA`/`onshell` share the code path but were not run.
- `fixed_order` is implemented (group dropped as a unit, sign on all members) but **unvalidated**.
- `nb_core > 1` throughout, so the additive shard merge is exercised; the serial path is not separately measured.

## Summary by Sourcery

Implement and validate MadSpin pure-interference samples with cross-helicity density restrictions, signed fully weighted events, and zero-cross-section bookkeeping.

New Features:
- Add a configurable MadSpin pure-interference mode that isolates interference between selected production and decay helicity sets and produces signed, fully weighted events.
- Record zero-cross-section diagnostics and reference normalisation details in the generated LHE banner.

Bug Fixes:
- Preserve the normalising density-matrix trace separately from cross restrictions so interference weights do not become undefined.
- Prevent malformed semicolon-separated pure-interference card entries from being silently ignored.

Enhancements:
- Support cross helicity restrictions that retain both index orderings, remain real, compose across particles, and preserve existing symmetric-restriction behaviour.
- Force joint unweighting and add signed-weight handling, normalisation measurement, shard-safe accounting, and zero-integral statistical checks for interference samples.
- Enable the required helicity frame handling and validate incompatible spin modes, production polarisation, decay targets, and auxiliary polarisation weights.

Documentation:
- Document the pure-interference design, constraints, implementation status, validation results, and known boundaries in the sequential MadSpin plan.

Tests:
- Add unit coverage for cross-restriction algebra, trace separation, syntax and validation, frame selection, normalisation, and mode interactions.
- Validate the mode end to end with parallel MadSpin runs, signed event weights, zero total cross-section, matching weight magnitudes, and unchanged output when the option is disabled.


oliviermattelaer:
Heads-up on a merge interaction with #361 (`_needs_frame_axis()` wiring, now open against `madspin_density`).

This PR adds a third clause to `_frame_boost`'s guard for the interference mode:

```python
if (self._beampol() is None and not self._production_polarization()
        and not self._pure_interference()):
```

#361 replaces that whole guard with `if not self._needs_frame_axis(): return None`, and `_needs_frame_axis()` does **not** consult `_pure_interference()` — it reads a different card option. So merging the two will conflict in two places:

1. `_frame_boost` — both sides rewrote the same guard line.
2. `_needs_frame_axis`'s docstring — this branch still carries the "NOT WIRED IN ON THIS BRANCH" paragraph that #361 deletes.

Resolution when both land: keep #361's version (`if not self._needs_frame_axis()`) and add a fourth clause inside the helper returning true for `_pure_interference()`. The interference restriction is a projection like the others, so it needs the `me_frame` axis for the same reason — dropping the clause would silently put it back on the lab axis.

PR #360, opened 2026-08-18
MadSpin: auto picks sequential when the production is polarised

oliviermattelaer:
Follow-up to #349. `unweighting = auto` currently resolves on the spinmode family and the decay multiplicity, so offshell at two decaying particles gives `joint` — which is between one and three **orders of magnitude** slower once the production carries a polarisation brace.

## Reproduced independently

`p p > t t~` LO, 500 production events, three separate generations, `spinmode=madspin` (offshell), `decay t > w+ b, w+ > e+ ve` + conjugate, seed 42, `nb_core 1`, defaults `Nevents_for_max_weight=75` / `nb_sigma=4.5`. Trials per accepted event:

| production | joint | sequential |
|---|---|---|
| `t t~` (unpolarised) | **3.31** (`auto`) | 6.14 |
| `t{+}t~{+}` | **111.76** | 9.10 |
| `t{+}t~{-}` | **161.85** | 8.43 |

34x and 49x worse under joint on the polarised samples — while unpolarised joint is the *better* choice by ~1.9x, which is exactly why the clause has to be conditional rather than a blanket change.

The 50000-event closure validation of all four polarised final states saw the joint column at 204–213 (like-helicity) and 5800–6300 (opposite-helicity) against 8.59 sequential. The gap **widens with statistics**: at 500 events `nb_sigma` is 4.5 over a 75-event scan, at 50k it is 5.30 over 110, so the bound the joint test must clear keeps growing while the bulk of the restricted weight distribution does not. Same mechanism, consistent numbers.

After the change: `t{+}t~{+}` **111.76 → 9.10** (12.3x), `t{+}t~{-}` **161.85 → 8.43** (19.2x).

## The clause

In `_auto_unweighting_mode()`, after the PA/onshell branch and before the multiplicity rule:

```python
if self._density_spinmode() and self._production_polarization():
    return 'sequential'
```

Gated on `_density_spinmode()` so a non-density spinmode never parses the proc card, and so a malformed brace cannot raise from a mode that ignores braces.

**A brace on a particle MadSpin does not decay still fires the clause.** Such a brace leaves the restriction handed to `DensityMatrix` empty and so cannot be what peaks the weight, but firing anyway is right for two reasons, both in the docstring: the cost asymmetry (a spurious `sequential` costs ~2x, a missed one costs 30–1500x), and the fact that the resolved mode must be identical at every call site — it names the `max_wgt_sequential_*` cache files and selects which bound the accept/reject tests against, and the decayed-pdg set is not available everywhere `_unweighting_mode` is called, so a clause consulting it could resolve two ways within one run.

An explicit `set unweighting joint` is still honoured.

## Nothing else moves — verified, not assumed

Exhaustive grid, base vs patched, over 4320 combinations (6 unweighting values x 7 spinmodes x 5 `_nb_decaying` x polarised x `fixed_order` x `_decay_groups` x `density_method`):

```
total combinations: 4320
changed: 6
  unw=auto spinmode=full    pol=1 fixed_order=0 groups=0 density=1 : joint -> sequential (nb 0,1,2)
  unw=auto spinmode=madspin pol=1 fixed_order=0 groups=0 density=1 : joint -> sequential (nb 0,1,2)
changed with pol=0: 0
```

Corroborated by runs: unpolarised `auto` still logs `joint (auto, 2 decaying particle(s))` at 3.31 trials/event; polarised `spinmode=onshell` + `auto` logs `sequential (auto, ...)` at 4.43.

**The grid is now a permanent test rather than a scratch script.** `TestUnweightingDecisionTable` (added by #346, which pins the decision table against rules restated independently of the implementation) gained `polarization=` on its stub, the new branch in `_reference_mode`, and polarisation as a seventh dimension in `_cases()`. So the before/after comparison runs on every test run.

## Documentation

Both the `unweighting` option comment and `_auto_unweighting_mode`'s docstring gained the polarised branch with the measured table, in the existing style. `two_stage` stays unadvertised — #344's wording of the schemes sentence is kept intact and only the new `auto` sentence is appended.

## Verification

`python3 tests/test_manager.py test_madspin -t0`: **235 OK on the base**, **240 OK here**.

## Negative result worth keeping

The over-maximum-weight asymmetry under offshell+`sequential` reproduces (per 500 events: unpolarised 0, `t{+}t~{+}` 6, `t{+}t~{-}` 9 — the same direction as the validation's 2 vs 82–91 per 50k; `onshell` emitted none). But one extra run settles it: `t{+}t~{+}` with `Nevents_for_max_weight=300`, `nb_sigma=5.3` gave **0 overflows** at 9.39 trials/event, against 6 at 9.10 with the 75/4.5 defaults. So it is a **max-weight-statistics problem, not a polarisation-specific one** — the existing knobs fix it for ~3% in trials, and the auto-scaling already grows both with `nevents`. Not touched; the diff stays at the `auto` clause.

## Caveats

- One measurement (`t{+}t~{+}` under joint) came from a run launched before the edit landed, whose log was truncated past the announcement line. The exhaustive grid independently proves base `auto` = `joint` for that configuration, so the row stands.
- The resolution of the merge onto the refactored `_auto_unweighting_mode` shape is a **reconstruction**: an accidental `git checkout` overwrote the first pass, and it was rebuilt deterministically from the base file and then diffed against it to confirm intent. Independently re-verified by the coordinator (clause inside `_auto_unweighting_mode`, `two_stage` text absent, #344's wording present, clean merge into `madspin_density`, 240 tests green).
- `tests/parallel_tests` not run. No physics-closure check that polarised `sequential` output matches `joint` — that identity is what `sequential_debug` and the comparator ordering tests cover, and this change touches only which scheme `auto` names, not the schemes themselves.

## Summary by Sourcery

Select sequential unweighting automatically for polarised density-based MadSpin production while retaining all existing explicit and non-polarised selection rules.

Bug Fixes:
- Make `unweighting = auto` choose sequential unweighting for density-based spinmodes when the production process is polarised, avoiding severe performance degradation from joint unweighting.

Enhancements:
- Preserve existing multiplicity-based and explicit unweighting decisions, including explicit requests for joint mode and existing PA/onshell behavior.

Documentation:
- Document the polarised-production auto-selection rule and its measured performance characteristics.

Tests:
- Extend MadSpin decision-table coverage across polarisation states and add regression tests for polarised, unpolarised, explicit, and gated mode-selection behavior.

Chores:
- Improve test stubs to detect missing borrowed decision helpers and use explicit case records for exhaustive decision-table validation.

### DIFF
--- a/MadSpin/interface_madspin.py
+++ b/MadSpin/interface_madspin.py
@@ -181,7 +181,7 @@ class MadSpinOptions(banner.ConfigFile):
                        "sequential_global_retry: as sequential, but a rejected decay redraws the virtualities too. "
                        "sequential_with_mass: one test per decaying particle with that particle's virtuality drawn *inside* its own accept/reject, so nothing is ever frozen and no stage has a conditional normalisation to divide out. Needs a per-particle mass draw, i.e. the PA spinmode; elsewhere it falls back to sequential. "
                        "two_stage, sequential and sequential_global_retry unweight the set of virtualities first; the first two then need a tabulated running-width factor, measured during the max-weight scan to ~0.5%, which is far inside the pole approximation these modes already assume; sequential_global_retry does without it at 2-3x the cost, and is meant as a cross-check rather than a default. "
-                       "auto: sequential under PA/onshell, where it was the fastest scheme at every decay multiplicity measured; offshell joint up to two decaying particles and sequential from three, since offshell every mass set costs a production reshuffle and a production density and below three decays there are not enough of them to save to pay for it.")
+                       "auto: sequential under PA/onshell, where it was the fastest scheme at every decay multiplicity measured; offshell joint up to two decaying particles and sequential from three, since offshell every mass set costs a production reshuffle and a production density and below three decays there are not enough of them to save to pay for it; but sequential at every multiplicity when the production process carries a polarisation brace, since restricting the convolution to a polarisation subspace peaks the joint weight far below the single bound the joint test has -- measured on `p p > t t~` with both tops decayed, 112 trials per accepted event under joint for `t{+}t~{+}` and 162 for `t{+}t~{-}` against 9.1 and 8.4 under sequential, where unpolarised joint takes 3.3 (and at 50000 events, where the max-weight bound is looser still, the polarised joint columns were 204-213 and 5800-6300). An explicit 'set unweighting joint' is still honoured.")
         self.add_param('sequential_decay', 'auto',
                        comment='DEPRECATED, use unweighting: True maps to sequential, False to joint.')
         self.auto_set.add('sequential_decay')
@@ -2894,10 +2894,11 @@ class MadSpinInterface(extended_cmd.Cmd):
         spinmodes reshuffle the whole production onto the mass set at once, so
         there they fall back to ``sequential``.
 
-        ``auto`` has two branches, one per spinmode family. They were measured
-        over the number of decaying particles n on `p p > w+ j` (n=1),
-        `p p > t t~` (2), `p p > t t~ z` (3) and `p p > t t~ t t~` (4), 50000
-        events each -- see MADSPIN_SEQUENTIAL_PLAN.md section 12.
+        ``auto`` has two branches, one per spinmode family, plus an override for
+        a polarised production. They were measured over the number of decaying
+        particles n on `p p > w+ j` (n=1), `p p > t t~` (2), `p p > t t~ z` (3)
+        and `p p > t t~ t t~` (4), 50000 events each -- see
+        MADSPIN_SEQUENTIAL_PLAN.md section 12.
 
         **PA/onshell -> ``sequential``, at every n.** It was the fastest of the
         three at all four multiplicities, by 1.2x at n=1 rising to 3.8x at n=4.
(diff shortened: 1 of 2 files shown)


oliviermattelaer:
Addressed the review comment about the test helpers passing many positional parameters (`case[:6]` slicing) — pushed as `c386b8277`, test file only, `MadSpin/interface_madspin.py` untouched.

**The case tuple is now a `collections.namedtuple`** — `_Case(spinmode, unweighting, nb_decaying, fixed_order, decay_groups, polarization, density_method)`. namedtuple rather than a dataclass because it stays a tuple, so `str(case)` still works as an assertion `msg` and nothing else had to learn a new type; `collections` was already imported. `_cases()` builds by keyword, `_reference_mode(cls, case)` reads named fields, a new `_case_stub(case)` builds the stub by keyword, and every `case[:6]` / `case[6]` is gone from the four test methods that had them.

`_Stub.__init__` is now **keyword-only**, which is what actually prevents the failure mode the reviewer described: a positional call raises `TypeError` instead of silently rebinding arguments.

**Evidence the assertions did not change.** For every generated case, dumped the tuple plus the values the assertions compare (`_reference_mode`, `_unweighting_mode`, `_sequential_active`, `_sequential_upfront`, sorted pool ladder), driving each revision through its own `_cases`/`_Stub`/`_reference_mode`:

```
before: n_cases=5184  n_unique=5184
after:  n_cases=5184  n_unique=5184
diff before.txt after.txt  ->  no output (byte-identical, order included)
sort|md5 both: 75d75fa6288c9e83c595d659bafdcfc3
```

5184 = 9 spinmodes x 6 unweighting x 6 multiplicities x 2 fixed_order x 2 decay_groups x 2 polarisations x 2 density_method.

**Also addressed the related fragility in `_borrow_decision_helpers`**, which this batch broke four separate times by adding a call inside `_unweighting_mode` to a method the stubs did not borrow. It now installs a `__getattr__` naming the culprit —

```
_Stub does not borrow MadSpinInterface._decaying_pdgs: add it to the list in
_borrow_decision_helpers, or define it on the stub
```

— instead of a bare `AttributeError`. It stays an `AttributeError`, so `getattr(self, x, default)` and `hasattr` are unaffected (verified explicitly, since one test deletes `_nb_decaying` and relies on the production `getattr` fallback). The fully self-maintaining version (transitive closure over the borrowed sources) was deliberately not attempted: it would pull in an unpredictable set of real methods and could mask attributes a stub withholds on purpose.

`./tests/test_manager.py test_madspin -t0`: 240 OK before, 240 OK after — same count, as expected for a refactor. Still merges cleanly into `madspin_density`.

One thing to watch: if another in-flight branch adds a positional `self._Stub(...)` in this class, the merge surfaces that as a `TypeError`. Loud and one line to fix, but it will surface.

PR #363, opened 2026-08-18
MadSpin: unweighted-up-to-a-sign interference, and a general weighted-output option

oliviermattelaer:
**Stacked on #351.** Two related options, both defaulting to current behaviour.

## 1. `pure_interference_output = weighted | unweighted` (default `weighted`)

An **unweighted-up-to-a-sign** interference sample: unweight on `|W|` (one draw, accept on `|W|/M`, nothing written on rejection), carry the sign onto the event weight, and write `w = ± sigma*BR*<|W|>/c`. The file then has exactly **two** weight magnitudes.

The derivation, re-done independently: accepting with probability `|W|/M` gives `N_file = N*<|W|>/M`, and since `|W|*sign(W) = W`,

```
(1/N_file) * sum_written w*O  =  w0 * <W O> / <|W|>        <- M cancels
```

Matching `sigma*BR*<W O>/c` gives `w0 = sigma*BR*<|W|>/c`, with **no `max_weight`**. `mean(w) = 0` since `<W> = 0`.

Worth recording: the earlier `w = ± sigma*BR*maxwgt/c` formula in the design notes is **not wrong physics** — it is the same estimator normalised per event *read* rather than per event *in the file*, since `maxwgt/c = (<|W|>/c) * (N_read/N_file)`. But an LHE carries no `N_read`, and `IDWTUP = -4` means the mean over the file, so a consumer dividing by `N_file` would be off by `M/<|W|>` — a factor 13 here. That is what made `maxwgt` look load-bearing.

### The trap the algebra does not warn you about

**`<|W|>` must not be taken from the max-weight probe.** Unlike `c` it is *not* a decay-side constant (that is exactly the point of plan §13.7b), so the probe's ~110 production events do not pin it down and its 400 draws per event are correlated through that point's `|W|` scale. Measured: probe `3.168e-11 ± 0.46%` against a truth of `2.895e-11` — **9.4% high**, and 5.0% when blocked by production event. Not an ordering artefact: random 110-event subsamples show a 9.5% spread. A 9.4% error on `w0` is 9.4% on every physics number the sample produces.

The fix is exact and free: use the run's own `<|W|> = (N_file/N_drawn)*M`, which makes `N_file` cancel out of the estimator entirely.

### `M` cancels — demonstrated

`p p > t t~`, 50k events, `spinmode onshell`, the `(I,I)` block, one parent file, three bounds spanning 5.8x via `nb_sigma`:

| | weighted | M | 1.65M | 5.81M |
|---|---|---|---|---|
| events written | 50000 | 3764 | 2305 | 620 |
| distinct `\|w\|` | 49985 | **1** | **1** | **1** |
| +/− | 24921/25079 | 1855/1909 | 1135/1170 | 326/294 |
| `z` | −0.452 | −0.880 | −0.729 | +1.285 |
| `<C_nn>` | +0.037286±0.000366 | +0.036567±0.000859 | +0.037049±0.001111 | +0.035850±0.002011 |

**Closure**: `<C_nn>` committed +0.03657±0.00059 (v2) and +0.03626±0.00090 (v1); this run **+0.036567±0.000859**, pulls **−0.00 / +0.25**. `<C_rr>` +0.002458±0.000830, pulls +1.34 / −0.01.

**Variance penalty 5.5–6.2x** on identical events (5.5x `<C_nn>`, 5.5x `<cos phi_ll>`, 6.2x `<Delta phi>`) — consistent with the 5.8/5.7/6.1 measured from the other direction, which is why `weighted` stays the default.

## 2. `decay_output = weighted | unweighted` (default `unweighted`)

Skips the accept/reject for the ordinary density modes. Density spinmodes only — `madspin_v1`, `onshell_v1` and `none` raise; under `pure_interference` it warns and steps aside.

The normalisation falls out: `w = sigma*BR*W/c` gives `mean(w) = sigma*BR*<W>/c = sigma*BR`, exactly MG5's `IDWTUP = -4` convention, so `XSECUP` keeps its ordinary value. Measured rather than assumed: `mean(w) = 23.783611 ± 0.029563` against `sigma*BR = 23.779781` — **ratio 1.000161, 0.13 sigma**.

**4.13 decay trials per event -> 1**, for 6–13% more variance per event, i.e. **~3.7–3.9x less variance per unit CPU**. Note this is a *different and smaller* effect than the interference mode's 6x: there the accept/reject discards production events, here it only redraws decays, so weighted output is strictly noisier per event and the entire win is CPU.

## Caveats, stated plainly

- **A regression was introduced in feature 2 and caught by re-running, not by the tests**: the unweighted interference path fell through to a second, signed accept/reject (356 events instead of 3764, `z = +18.9`). Fixed, output byte-identical again, and six new tests now drive the real `_unweight_range` and fail on the bad condition. The mode's own `z` check did catch it — but only after a full run.
- **No downstream consumer was checked** for `decay_output = weighted`. Pythia8, Delphes and the rest were not run. The output is legal LHEF, but anything counting events rather than summing weights will be wrong. This is said in the option comment and in the banner.
- `<|W|>`-from-the-run is exact in expectation, but `N_file` is random, so `w0` carries a `1/sqrt(N_file)` self-normalisation error (1.6% here) — small next to the 5.5x variance penalty.
- Only `spinmode = onshell` was exercised end to end for both new paths. `PA` — the one place the `W <= 0` failed-reshuffle handling can fire — and `fixed_order` are implemented but unvalidated.

## Verification

Byte-identity with both options at their defaults: same 90,368,979-byte file, SHA-256 `767da240…f2855`, against the base branch. `test_madspin -t0`: **325 tests, OK** (291 on the base). Plan §13.17 and §13.18 record both features.

## Summary by Sourcery

Add optional weighted decay output and signed-unweighted pure-interference output while preserving MadSpin's existing defaults.

New Features:
- Add configurable pure-interference output that supports signed, constant-magnitude event weights through |W|-based unweighting.
- Add configurable weighted decay output for density spin modes, retaining one decay draw per production event with the convolution carried by event weights.

Bug Fixes:
- Prevent pure-interference unweighted events from undergoing a second signed accept/reject decision.
- Normalize unweighted interference weights using the run-realized keep rate instead of the maximum-weight probe's potentially biased absolute-convolution estimate.
- Handle failed non-interference reshuffles as zero-weight events in weighted decay output.

Enhancements:
- Extend normalization scans, caching, accounting, LHE banner reporting, and output metadata for weighted and signed-unweighted modes.
- Preserve existing default behavior and byte-identical output when the new options are not enabled.

Documentation:
- Document the estimator normalization, supported modes, performance trade-offs, validation results, and caveats for both output options in the MadSpin sequential plan.

Tests:
- Add unit coverage for option validation, normalization estimates and caches, estimator behavior across bounds, weighted decay accounting, event-weight rescaling, and all output paths through the unweighting loop.

### DIFF
--- a/MadSpin/interface_madspin.py
+++ b/MadSpin/interface_madspin.py
@@ -3900,9 +3900,12 @@ class MadSpinInterface(extended_cmd.Cmd):
         pure_interference_c = ctx.get('pure_interference_c')
         pi_unweighted = pure_interference and bool(
             ctx.get('pure_interference_unweighted'))
-        # every trial is kept and carries W/c
-        keep_every_trial = weighted_decay or (pure_interference
-                                              and not pi_unweighted)
+        # No ordinary accept/reject is made below in any of these modes: the
+        # two weighted paths keep every trial outright, and the 'unweighted'
+        # interference path has ALREADY made its own decision (on |W|, one
+        # draw) by the time the test is reached, so falling through to the
+        # signed test there would reject every negative weight a second time.
+        no_joint_test = pure_interference or weighted_decay
         if (pure_interference or weighted_decay) and not pure_interference_c:
             raise self.InvalidCmd(
                 "MadSpin: the normalisation constant c = <W> is missing; the "
@@ -4121,7 +4124,7 @@ class MadSpinInterface(extended_cmd.Cmd):
                     dead_trials = self._dead_trial(dead_trials, wgt,
                                                    'the joint accept/reject')
 
-                if keep_every_trial or random.random()*maxwgt < test:
+                if no_joint_test or random.random()*maxwgt < test:
                     if offshell_density:
                         # prod_trial has already been reshuffled internally (its
                         # jacobian is in wgt); build the event to write out from the
--- a/MadSpin/interface_madspin.py
+++ b/MadSpin/interface_madspin.py
@@ -160,6 +160,36 @@ class MadSpinOptions(banner.ConfigFile):
         self.add_param('global_order_coupling', '')
         self.add_param('identical_particle_in_prod_and_decay', 'average')
         self.add_param('beampol', [0., 0.], comment='beam polarisation of each beam in percent, -100 .. 100, exactly as the run_card polbeam1/polbeam2 (0 is unpolarised). Taken from the run_card of the production when it has one.')
+        self.add_param('decay_output', 'unweighted',
+                       allowed=['unweighted', 'weighted'],
+                       comment="whether MadSpin unweights its decays at all. 'unweighted' "
+                       "(default, and what MadSpin has always done): draw decay "
+                       "configurations until one is accepted, so every production event "
+                       "yields exactly one output event and every event of a given "
+                       "production process carries the same weight. 'weighted': NO "
+                       "accept/reject -- draw ONE decay configuration per production event, "
+                       "keep it, and put the convolution on the weight, "
+                       "w = w_prod * BR * W / c, with W this trial's production/decay "
+                       "density convolution and c = <W> its decay-phase-space mean (a "
+                       "constant, measured by the same probe that estimates the maximum "
+                       "weight). Under MG5's IDWTUP = -4 convention the cross-section is "
+                       "the MEAN of the weights, and mean(w) = sigma*BR exactly as before, "
+                       "so <init> is unchanged and the file is self-normalising. WHY, "
+                       "measured on p p > t t~ with spinmode = onshell and 50 000 events: "
+                       "the ordinary accept/reject burned 4.13 decay trials per written "
+                       "event, each one a matrix-element evaluation, and this mode burns "
+                       "exactly 1; the price is only a 6-13% larger variance per "
+                       "production event (the weights have sd/mean = 0.28), so per unit "
+                       "of CPU in the unweighting loop it is about a 3.7-3.9x variance "
+                       "reduction. WHAT YOU GIVE UP: the output is a WEIGHTED LHE "
+                       "file. Anything downstream that assumes MadSpin events carry a "
+                       "constant weight -- unit-weight event counting, simple histogram "
+                       "entry counts -- is wrong on it. Density spin modes only "
+                       "(madspin/full, PA, onshell): the v1 modes and spinmode = none "
+                       "build no density matrix and have no W. Ignored under "
+                       "pure_interference, which is always weighted and has its own "
+                       "pure_interference_output. See doc/madspin_sequential_plan.md "
+                       "section 13.18.")
         self.add_param('pure_interference', '',
                        comment="pure-interference mode: keep ONLY the interference between two "
                        "polarisations of a decaying particle in the production/decay density "
@@ -1641,6 +1671,9 @@ class MadSpinInterface(extended_cmd.Cmd):
             self.options['spinmode'] = spinmode
 
         logger.info("Running MadSpin in spinmode %s" % spinmode)
+        # decay_output is refused outside the density modes too, so it is
+        # checked before the branch rather than inside it
+        self._validate_weighted_decay()
         if self._density_spinmode():
             # read (and validate) the production polarisation braces now rather
             # than on the first event, deep inside a worker process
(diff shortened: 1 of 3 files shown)


oliviermattelaer:
## PA and madspin validation — the caveat in this PR is now closed

The PR notes that only `spinmode = onshell` was exercised end to end for both new output paths. Both have now been validated under **PA** and **madspin** as well. Full write-up, scripts and raw numbers on `claude/ms-interference-pa-validation` (`MadSpin/validation/interference_pa_madspin/`); no shipped source is touched by that branch.

### The physics agrees across all three spinmodes

50 000 production events of `p p > t t~`, one shared production sample decayed six ways, the `(I,I)` block. `<C_nn>` against the committed `+0.03657 +/- 0.00059`:

| | weighted | unweighted |
|---|---|---|
| onshell (control) | `+0.036964 +/- 0.000322` (+0.59) | `+0.037828 +/- 0.000644` (+1.44) |
| **PA** | `+0.037353 +/- 0.000335` (**+1.15**) | `+0.036016 +/- 0.000725` (**-0.59**) |
| **madspin** | `+0.037017 +/- 0.000326` (**+0.66**) | `+0.036320 +/- 0.000653` (**-0.28**) |

`<C_rr>` pulls +0.31 to +1.66; `<C_kk>` consistent with zero throughout. Zero-cross-section holds everywhere: `|z| <= 1.92`, `mean(w)` compatible with zero, `XSECUP = 0`, 0 dead trials, 0 overweights, both signs present. The `unweighted` output holds **exactly two signed weight values and one magnitude** in all three spinmodes.

`c` measured vs analytic: **0.9974 (onshell) against 0.9554 (PA) / 0.9563 (madspin)** — `<jac> ~ 0.955`, 33 sigma from 1. The code correctly measures rather than assumes, which is exactly what the offshell Breit-Wigner jacobian demands. The probe's `<|W|>` was 8.2% (PA) / 8.7% (madspin) high and was correctly replaced by the run's own value.

### The `W <= 0` path: never fires, so it was forced

**0 non-positive returns and 0 internal retries in 376 064 `reshuffle_production` calls.** `_draw_mass_value` caps each virtuality at the remaining `sqrt(shat)` budget, so RAMBO's `-1` branch is unreachable from that direction. The sweep therefore does **not** test the handling, and fault injection was used instead (10% of trials, event loop only, against a `p=0` control):

- **`jac = 0`, weighted** -> written with `w = 0` exactly (200 injected, 200 zeros), `z` unmoved. Correct.
- **`jac = 0`, unweighted** -> **zero** reach the file; `random()*M >= 0` always rejects. Correct.
- **`jac = -1`, either output** -> **every monitor stays silent** (`z` +0.44 / -0.66, dead counter 0) while `<C_nn>` drops to 0.71 / 0.67 of the control (-4.7 / -2.4 sigma). In the unweighted case the invalid trials are accepted *more* often than real ones (146 vs 116 kept).

**This is not a live bug**: `jac = -1` cannot reach either path, because both call `reshuffle_production(_allow_retry=True)`, which resamples and recurses on `jac in (0, -1)`; the `_allow_retry=False` callers belong to the sequential schemes, which `pure_interference` forces off. Only `jac = 0` can reach the caller, and both paths handle it correctly.

Worth recording as a fragility rather than a defect: the protection lives in `lhe_parser`, not in the mode, and the `z` test structurally cannot see a sign flip. `if signed <= 0 and jac <= 0` would close it and is a no-op today. Related: `_dead_trial` does not run on the interference path at all on this base, so `nb_pi_dead` (non-finite only) is the whole monitor.

### `fixed_order` — two pre-existing defects, not from this PR

Exercised with a synthetic `<eventgroup>` file, with `pure_interference` **off**:

1. **Counter-events are never decayed, in every spinmode** — `add_decays(decays)` is called after the born event already drained the dict (`Event.add_decays` shallow-copies the dict and `pop(0)`s the caller's lists).
2. **Under PA not even the born event is decayed** — `fixed_order` forces `build_event=True`, `get_onshell_evt_and_wgt` consumes `decays`, and the PA `density_keep_jacobian` branch rebuilds from the empty dict.

Measured as 4 particles/member (undecayed) against 12 (decayed). The responsible line comes from `1bb7ee1de` on `madspin_density`, an ancestor of this stack's base, and this PR's own bookkeeping on top of it is correct. Recommend `fixed_order` not be advertised with the density spinmodes until fixed — separate change, not a blocker here.

PR #366, opened 2026-08-19
MadSpin: archive the card that actually ran, including under ms_dir

oliviermattelaer:
**Stacked on #365**, which fixed the sibling bug in the same `curr_dir`/`path_me` family and introduced the single-accessor pattern followed here. Composes cleanly with #364.

`do_launch` archived the MadSpin card from `pjoin(curr_dir, 'Cards', 'madspin_card.dat')`. Under an `ms_dir` run `curr_dir` can point inside the gridpack, where no `Cards/` directory exists — and the block is guarded by `if os.path.exists(...)`, so the archive is skipped **in silence**. No warning, no log line (grepped a full run log for `madspin_card`/`RunMaterial`: zero hits).

## Sharper than first reported: it depends on card ordering

Not "any `ms_dir` run". The deciding factor is whether a `do_import` runs *after* `set ms_dir`. Reproduced end to end on the base (`p p > t t~`, both tops decayed, `spinmode madspin_v1`):

| how MadSpin was driven | `curr_dir` at archive time | archived card |
|---|---|---|
| MadEvent ordering (constructor imports events), no `ms_dir` | process dir | **present** |
| MadEvent ordering, card says `set ms_dir` | the gridpack | **absent** |
| card: `import` … then `set ms_dir` | the gridpack | **absent** |
| card: `set ms_dir` … then `import` | process dir | **present** |
| second run reusing the gridpack (`run_from_pickle`) | the gridpack | **absent** (never archived at all) |

The consequence that matters: **the losing ordering is the one MadEvent always produces.** `common_run_interface.do_decay_events` constructs `MadSpinInterface(args[0])`, which imports the event file in `__init__`, and only then runs `Cards/madspin_card.dat`. So every MadEvent run whose madspin card contains `set ms_dir` lost its card. Standalone cards lose it only when they import before setting `ms_dir`.

## Root cause and fix

`curr_dir` means "where this run's output goes" — its meaning at every other use site, and what #365 pinned it to for the decayed events. `post_set_ms_dir` re-points it at the gridpack; `do_import` points it back. Deriving the *card's* location from it was the same category error as #365, invisible whenever the two coincide.

Following #365, the source is now a single accessor, `MadSpinInterface.madspin_card_path`, answering "which card did this run execute": the file handed to `import_command_file` (recorded by a thin override — literally the card the user edited, and it need not live anywhere in particular), falling back to `Cards/madspin_card.dat` under `event_base_dir`, the directory `do_import` derived from the event file and now records under its own name so a later `set ms_dir` cannot move it. Nothing in the chain reads `curr_dir`.

The copying moved into `_archive_madspin_card(decayed_evt_file)`, called from `do_launch` **and** `run_from_pickle` — so the gridpack-reuse path, which archived nothing at all before, now archives the same file from the same source.

## Neighbouring archiving — surveyed

- `run_from_pickle` (every rerun against an existing `ms_dir`) had **no** archiving code. Confirmed by running the base twice against one gridpack. Fixed here.
- `run_bridge` (`spinmode none`) and `run_onshell` (`onshell`, `onshell_v1`, `PA`, and the density `madspin`/`full`) never archive the card. **Left unchanged, but worth flagging**: in upstream 3.x the default `madspin`/`full` fell through to `do_launch` and *did* archive; this fork routes it to `run_onshell`, so the default mode quietly stopped keeping the card. That is a wider behaviour change than this fix, so it is reported rather than made, and tracked separately.
- Pre-existing quirk, unchanged: the archive lands in `dirname(decayed_evt_file)`, the *production* run directory, while `do_decay_events` afterwards moves only the LHE into the new `<run>_decayed_N` directory — so the card copy stays behind with the production run.
- Mitigating context: in the MadEvent flow the card also survives in the decayed run's banner (`<madspin>` block via `banner.add(path)`, verified on an existing decayed run). The loss is total only for standalone MadSpin.

## Verification

- After the fix all four orderings archive, and the archived bytes are `diff`-identical to the card that was run; the gridpack-reuse run adds `..._decayed_1.dat` rather than overwriting.
- The **non-`ms_dir`** path still archives correctly — the case that breaks if the fix is made at the wrong end.
- `tests/test_manager.py test_madspin -t0`: baseline re-measured with `git stash` = **253 OK**; after = **268 OK** (new `TestMadSpinCardArchive`, 15 tests).
- The tests drive a real `MadSpinOptions` and the real `do_import`, so the `post_set_ms_dir` -> `do_import` interaction comes from production code rather than being simulated, and they tie the two orderings to each other rather than each to a literal.
- **Counterfactuals, all validated**: base interface + new tests -> 3 failures, 11 errors; a "fix" that shares the copying but still derives the card from `curr_dir` -> 8 failures/errors including the ordering pin; and the other tempting fix, stopping `post_set_ms_dir` from moving `curr_dir`, fails the premise test here **and** #365's `test_writer_and_reader_agree_when_ms_dir_is_curr_dir` — which is exactly why that fix is wrong.

## Not verified

The real `common_run_interface.do_decay_events` was not exercised end to end (no full MadEvent generation); its ordering was reproduced faithfully by constructing `MadSpinInterface(evt)` then `import_command_file(card)`, which is what that function does. The `RunMaterial.tar.gz` branch is covered by a unit test rather than a real aMC@NLO run.

## Summary by Sourcery

Ensure MadSpin consistently archives the card used to generate each decayed event, including runs that use or reuse an `ms_dir` gridpack.

Bug Fixes:
- Archive the MadSpin card that was actually executed across `ms_dir` configurations, card-ordering variations, and gridpack-reuse runs.
- Preserve archived MadSpin cards inside `RunMaterial.tar.gz` when applicable and avoid overwriting cards from earlier runs.

Enhancements:
- Centralize MadSpin card discovery and archiving so the source is based on the executed card and event location rather than the mutable output directory.
- Add comprehensive regression coverage for card ordering, alternate card locations, interactive sessions, repeated runs, and packed run material.

Tests:
- Add unit tests covering MadSpin card discovery, archival behavior, reruns, and RunMaterial packaging.

### DIFF
--- a/MadSpin/interface_madspin.py
+++ b/MadSpin/interface_madspin.py
@@ -623,7 +623,12 @@ class MadSpinInterface(extended_cmd.Cmd):
         self.err_branching_ratio = 0
         self.me_run_name = "" # Events diretory name where to stotre the events (used by madevent) not use internally
         self.all_iden = {}
-        
+        # The card this run executes, and the directory do_import derives from
+        # the event file. Both are what madspin_card_path archives from; set
+        # before the do_import below, which fills the second one in.
+        self.ms_card_path = None
+        self.event_base_dir = None
+
         if event_path:
             logger.info("Extracting the banner ...")
             self.do_import(event_path)
@@ -731,9 +736,16 @@ class MadSpinInterface(extended_cmd.Cmd):
         # change directory where to write the output
         self.options['curr_dir'] = os.path.realpath(os.path.dirname(inputfile))
         if os.path.basename(os.path.dirname(os.path.dirname(inputfile))) == 'Events':
-            self.options['curr_dir'] = pjoin(self.options['curr_dir'], 
+            self.options['curr_dir'] = pjoin(self.options['curr_dir'],
                                                       os.path.pardir, os.pardir)
-        
+        # Keep that directory -- the process root when the events sit in
+        # Events/<run>/, the event file's own directory otherwise -- under a
+        # name of its own. 'set ms_dir' re-points curr_dir at the gridpack
+        # (post_set_ms_dir), so curr_dir stops answering "where did these
+        # events come from" as soon as a card mentions ms_dir after the import.
+        # See madspin_card_path.
+        self.event_base_dir = self.options['curr_dir']
+
         if not os.path.exists(inputfile):
             if inputfile.endswith('.gz'):
                 if not os.path.exists(inputfile[:-3]):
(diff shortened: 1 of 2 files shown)


oliviermattelaer:
Full unit suite check, for the record: `tests/test_manager.py -t0` (1106 tests) reports **3 errors** with this change — `test_wj_loonly_gen`, `test_DensityMatrixObservables22`, `test_fks_ppzz_in_RS`. None are related to card archiving.

Re-run in isolation on this tree and on the base commit `b939412b5`: **identical results on both**. `test_DensityMatrixObservables22` and `test_fks_ppzz_in_RS` error the same way on each (the visible traceback is a UFO model-loading failure, `models.UFOError: 'dict' object has no attribute 'iteritems'` — a Python-2-ism in a model file); `test_wj_loonly_gen` passes in isolation on both, so its full-suite failure is a cross-test state effect rather than anything this branch does.

The targeted suite is the meaningful one here: `test_madspin -t0` goes **253 on base → 268 with the change**, all green.

oliviermattelaer:
Full-suite comparison now complete, confirming the earlier note:

- **with this change**: 1106 tests, 3 errors
- **on the base** (`b939412b5`): 1091 tests, the **same 3 errors**

Identical failure set — `test_wj_loonly_gen`, `test_DensityMatrixObservables22`, `test_fks_ppzz_in_RS` — with the visible cause being a Python-2-ism in a UFO model file (`models.UFOError: 'dict' object has no attribute 'iteritems'`). Pre-existing and unrelated to card archiving.

The 15-test delta between the two runs is exactly the new `TestMadSpinCardArchive` class.

PR #377, opened 2026-08-19
MadSpin: bound the mass stage per production event, for any 2 -> N

oliviermattelaer:
**Stacked on #375** (`c8b79c0e9`), whose overflow counters are the instrument that
shows this bound actually dominates. Review #375 first. It does not yet contain
#375's later negative-weight commit `d0019d4b8`; rebase before merge.

The sequential mass stage used a single global bound `C`, taken from a probe over
the first `Nevents_for_max_weight` events and scaled by `nb_sigma`. It has to
dominate every event in the run, so it is set by the worst event in the sample
and every ordinary event pays for it.

This makes the bound per production event.

## Why it works for any 2 -> N

`Event.mass_shuffle` is RAMBO eqs. (4.2)/(4.3)/(4.9). Its only dependence on the
production event is the CM-frame `{E_i, m_i^2}` - n numbers, computed once. Given
those, a jacobian evaluation is one scalar Newton solve plus O(n) arithmetic. The
`2 -> 2` closed form is just the case where the Newton solve is explicit; it was
never what made this affordable.

What made it expensive was the machinery around it in `_production_jacobian_for`:
`lhe_parser.Event(str(production))` re-parses the event from a string, then
`split_event_by_onshell_propagator`, `FourMomentum` construction, a Lorentz map
per particle, the momentum rebuild and the boost back - none of which a bound
needs. This PR adds a pure kernel with none of that.

## Monotonicity is proved, not measured

With `a_i = |p_i|^2` and `beta_i = |p_i'|/E_i'`, eq. (4.9) is
`J = const . chi^(3n-5) / (G . prod E_i')`, and differentiating `sum E_i' = sqrt(s)`:

```
2 E_k' G chi^2 . dlnJ/dm_k'^2 = -(3n-5) + beta_k^2 + sum beta_i^2
                                - (sum E'beta^4)/(sum E'beta^2) - (sum E'beta^2)/E_k'
```

The third term is >= 0 and the fourth is >= `beta_k^2` (its own summand), leaving
`<= 5 - 2n < 0` for every n >= 3; n = 2 is the `lambda^(1/2)` ratio, decreasing in
each mass. So the maximum of `J` is at the low corner of the window, exactly, for
any n, with no scan. 1.6e6 adversarial directional probes (n = 2..10, |p| and m
over eight decades): zero violations.

## The bound

```
C_e = J(low corner) . prod_s jac_BW_s . prod_s max_m Zhat_s(m)
```

`max Zhat` is the exact endpoint-or-vertex maximum of the log-quadratic, not a
sample mean with a safety margin - a mean would clip on any process where
`Zhat(m)` has structure (see the `BW_cut 70` row below). Falls back to the global
bound for offshell spinmodes, status-2 productions, non-fitting windows and
infeasible corners.

## Validation

Kernel vs `_production_jacobian_for`, ~4800 comparisons per n:

| n | agreement | verdict cases | shipped | kernel | speedup |
|---|---|---|---|---|---|
| 2 | 8.9e-16 | 1919, all agree | 33.0 us | 5.3 us | 6x |
| 3 | 6.6e-16 | 1966, all agree | 40.3 us | 6.2 us | 7x |
| 4 | 6.7e-16 | 1921, all agree | 47.1 us | 6.9 us | 7x |
| 5 | 8.8e-16 | 1877, all agree | 55.0 us | 7.8 us | 7x |

Real runs, same sample / seed / `ms_dir`:

| process | eps_m | overweights |
|---|---|---|
| `p p > t t~` | 1.80 -> **1.39** | 20/20000 -> **0** |
| `p p > t t~ j` (2->3) | 2.76 -> **1.50** | 0/8000 -> 0 |
| `p p > t t~`, `BW_cut 70` | 1.74 -> **1.68** | 23/20000 -> **0** |

Offline probe over ~1e6 mass sets: overflows 4971 -> **0**, largest
`max(w)/C_e` anywhere 0.9635.

The third row is the non-smooth-`Zhat` test: `BW_cut = 70` puts the top's `bW`
threshold at 85 GeV inside the window, and the fitted table runs
`Z(70.3)=0.002, Z(173)=1, Z(275.5)=0.263` - a factor 500 across the window. There
the bound barely tightens, but it is a bound: 23 overflows -> 0.

**The accepted distribution is unchanged**, which is the safety case: the accepted
density is `q_e(m).min(1, w/C)`, so any `C >= max w` cancels. Verified rather than
asserted - chi2/dof 0.58-1.25, KS p 0.22-0.99 across six comparisons.

`tests/test_manager.py test_madspin -t0`: **408 tests, OK** (was 395).
`test_lhe_parser`: 21, OK.

## Behaviour change

`nb_sigma` / `Nevents_for_max_weight` no longer set the *mass* bound (they still
set the angle bounds). This is what finally makes `eps_m` reproducible - the
+-40% run-to-run scatter recorded in `doc/madspin_pa_mass_stage/findings.md` was
exactly that knob dependence. Logged, not silent.

## Not covered

Offshell spinmodes (`madspin`/`full`) keep the global bound - `Tr(rho_off)/|M|^2_on`
has no cheap maximum. Status-2 productions fall back. Not tested: more than two
decaying slots, the fixed-order path, multi-core. The non-smooth-`Zhat` case
needed a widened `BW_cut`; no default-`BW_cut` SM process was found with a
daughter threshold inside the window.

## Summary by Sourcery

Replace the global sequential mass-stage bound with a proven, efficient per-production-event bound for supported PA/onshell 2→N processes, while retaining global fallbacks for unsupported configurations.

New Features:
- Bound the sequential mass stage separately for each production event across general 2→N processes.
- Add a fast RAMBO mass-shuffle Jacobian kernel and exact Breit–Wigner rate-factor maxima for constructing per-event bounds.
- Report per-event versus global mass-bound usage and fallback counts.

Bug Fixes:
- Prevent mass-stage overweight overflows by using proven per-event upper bounds where supported.
- Ensure offshell, status-2, and infeasible cases correctly fall back to the global bound.

Enhancements:
- Make mass-stage efficiency reproducible by decoupling its bound from nb_sigma and Nevents_for_max_weight while retaining those controls for angle bounds.
- Preserve the accepted mass distribution while reducing unnecessary rejection overhead.

Documentation:
- Document the per-event mass-bound construction, its applicability, validation, and fallback behavior.

Tests:
- Add kernel equivalence, monotonicity, bound-dominance, fallback, caching, event-counting, and distribution-preservation tests for the mass stage.

### DIFF
--- a/MadSpin/interface_madspin.py
+++ b/MadSpin/interface_madspin.py
@@ -8292,7 +8292,15 @@ class MadSpinInterface(extended_cmd.Cmd):
     # What is NOT bounded here, and falls back to the global probe bound:
     #  * the offshell spinmodes (madspin/full), whose w_mass carries the extra
     #    Tr(rho_off)/|M_prod|^2_on -- a matrix-element ratio with no cheap
-    #    maximum. Only PA/onshell is covered;
+    #    maximum. Only PA/onshell is covered. That ratio is not why: measured on
+    #    3.9e6 free mass sets it is 1.0000 +- 0.012. What kills the construction
+    #    offshell is Zhat, whose window spans a factor 3.2 there against 1.16
+    #    under PA, so max_m Zhat sits 2.9x above the typical weight and the
+    #    corner bound comes out LOOSER than the probe's global one (eps_m 5.00
+    #    against 3.06). Even the partial bound J_corner . max_sample(rest) is
+    #    worth 2%, and is a loss once the run-level factor is measured as a
+    #    maximum rather than extrapolated. Section 15 of
+    #    doc/madspin_sequential_plan.md has the numbers;
     #  * a production event carrying an onshell propagator (status 2), where
     #    reshuffle_production multiplies in a reshuffle_decay jacobian per
     #    sub-decay that is not part of this factorisation;
@@ -8333,8 +8341,10 @@ class MadSpinInterface(extended_cmd.Cmd):
             if getattr(self, '_mass_bound_fallback_announced', False):
                 return
             self._mass_bound_fallback_announced = True
-            reason = ('the spinmode is not PA/onshell, so the mass weight '
-                      'carries an offshell production matrix element'
+            reason = ('the spinmode is offshell (madspin/full), where the '
+                      'per-event construction was MEASURED to be looser than '
+                      'this bound, not merely unavailable -- see section 15 of '
+                      'doc/madspin_sequential_plan.md'
                       if offshell else self._MASS_BOUND_UNSUPPORTED)
             logger.info(
                 "MadSpin sequential: the mass stage keeps the probe's global "
@@ -8857,10 +8867,17 @@ class MadSpinInterface(extended_cmd.Cmd):
         # cancels out of it. Changing C therefore changes the trial sequence and
         # the cost, and nothing about the sample.
         mass_bound = None
-        if probe is None and maxwgts and draw_mass and not offshell:
+        if probe is None and maxwgts and upfront and draw_mass and not offshell:
             mass_bound = self._mass_stage_bound(production, order, particles,
                                                 slot_to_index, zkeys, keep_jac)
-        if maxwgts and draw_mass:
+        # Which events *have* a mass stage to bound: the up-front schemes, and
+        # there whichever family draws a virtuality up front. ``_upfront_production``
+        # fills slot_mass under `offshell or draw_mass`, and `draw_mass` alone is
+        # the PA half of that -- so gating on it left the offshell spinmodes
+        # counted in neither column and never announced, and let
+        # sequential_with_mass (which is not an up-front scheme at all, and whose
+        # mass_bound is dead) announce a bound it does not use.
+        if maxwgts and upfront and (offshell or draw_mass):
             # one per chain call, i.e. one per production event reaching the
             # mass stage -- a rejected mass set loops *inside* the chain, so
             # these count events and not draws
--- a/MadSpin/interface_madspin.py
+++ b/MadSpin/interface_madspin.py
@@ -4666,6 +4666,15 @@ class MadSpinInterface(extended_cmd.Cmd):
                         drawn, rejects,
                         ', %d dropped by a rejected decay' % exact_restarts
                         if exact_restarts else '')
+        per_event = merged.get('nb_mass_bound_event', 0)
+        global_bound = merged.get('nb_mass_bound_global', 0)
+        if per_event or global_bound:
+            logger.info(
+                "MadSpin sequential mass stage: %d/%d production events used "
+                "the per-event bound%s", per_event, per_event + global_bound,
+                '' if not global_bound else
+                ' (%d fell back to the probe\'s global maximum weight)'
+                % global_bound)
         positions = sorted(int(k.rsplit('_', 1)[1]) for k in merged
                            if k.startswith('nb_try_'))
         for position in positions:
@@ -7960,11 +7969,16 @@ class MadSpinInterface(extended_cmd.Cmd):
                                 ncomb=len(hel), dimension=len(hel),
                                 frame_boost=frame_boost, frame_rest_leg=rest_leg)
 
-    def _draw_mass_value(self, pdg, budget):
-        """Sample one resonance virtuality from its Breit-Wigner, capped at the
-        remaining ``budget`` (what is left of sqrt(shat)). Returns
-        ``(mass, reshuffle_info, jac_bw)`` where jac_bw is the Breit-Wigner
-        sampling jacobian (gap/pi)."""
+    def _mass_window(self, pdg, budget):
+        """The Breit-Wigner sampling window of one resonance and its sampling
+        jacobian: ``(pole, width, min_mass, max_mass, jac_bw)``.
+
+        Note ``jac_bw`` is a function of the *window*, not of the mass drawn in
+        it: the sampler is uniform in R = atan((m^2-pole^2)/(pole.Gamma)) and
+        gap/pi is exactly that window's width in R over pi. That is what makes
+        the mass-stage bound of ``_mass_stage_bound`` a maximum over the window
+        corners rather than a scan in m.
+        """
         pole = self.banner.get('param', 'mass', abs(pdg)).value
         width = self.banner.get('param', 'decay', abs(pdg)).value
         if self.options['BW_cut'] < 0:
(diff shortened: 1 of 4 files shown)


oliviermattelaer:
Pushed `d2253d37a` on top: the offshell modes were **measured** for a per-event
bound and deliberately keep the global one, plus a reporting gap in this PR that
the measurement exposed.

**The measurement.** 3.90e6 free mass sets through `_upfront_production` on
`p p > t t~`, 10 000 events, `spinmode madspin`. The estimator reproduces the
runs' own logs (predicts 3.06 offshell, run reports 3.06; predicts 1.400 for the
shipped PA bound, run reports 1.41).

| bound | eps_m |
|---|---|
| global - shipped | **3.06** |
| `J_corner . combine(max R.jac_BW.Zhat)` | **3.00** |
| same, run-level factor as an honest maximum | 3.26 |
| fully factorised per-event | 5.00 |
| per-event supremum of w (unreachable) | 1.42 |

2 % - and worse than global below 400 GeV. Not taken.

**Why, and it is not the offshell ratio.** Over 3.90e6 mass sets
`R = Tr(rho_off)/|M_prod|^2_on` is `1.00000 +- 0.0119`, range [0.733, 1.314] - the
flattest factor in the weight. The gap is elsewhere:

- `Zhat` is steep offshell and flat under PA: `Z(150.6)/Z(173)/Z(195.3)` is
  `0.522/1/1.699` offshell against `0.912/1/1.059` under PA. The corner
  construction multiplies `max jac_BW` (low end of the window) by `max Zhat`
  (high end), which costs 7 % under PA and 190 % offshell - the whole gap between
  PA's 1.40 and offshell's 5.00.
- `J_corner` and the rest are anti-correlated across events (`corr = -0.49`), both
  driven by sqrt(shat). A maximum *of the product* sees that; a product of maxima
  does not. Fully factorising costs 5.7x here.

A tabulated `max R` was assessed and not taken: a 7x7 table in `(ln m1, ln m2)`
runs 1.02-1.31 against a single global 1.31, i.e. at most 25 % of a factor that is
already 1, for a probe record per mass set and a cache-format bump.

**The reporting gap this PR shipped.** The new counters and the bound
announcement were gated on `draw_mass`, the PA half of the condition
`_upfront_production` fills `slot_mass` under (`offshell or draw_mass`). So an
offshell run appeared in neither column and never said which bound its mass stage
used, and `sequential_with_mass` - not an up-front scheme, `mass_bound` dead there
- announced a bound it does not use. Both fixed.

**Validation.** 409 tests OK; the new test fails on `87195c33a` and passes here.
10 000-event `madspin` runs before/after with a warm `ms_dir`: decayed LHE
byte-identical, eps_m 3.07 both. The bound-cancels argument shown on the offshell
path by tripling `maxwgts[0]` in the cache: eps_m 3.07 -> 8.99 (3x, as it must),
overweight counters 2 -> 0, accepted virtualities unchanged (chi2/dof 0.72 and
1.14, KS p 0.98 and 0.56).

One correction to the PR's own text: `spinmode = full` is not a separate path,
`do_launch` aliases it onto `madspin` unconditionally.

oliviermattelaer:
Re-measured on this branch (`d2253d37a`): the efficiency table at 100 000 events
and an 18-cell lineshape campaign at 200 000, both reusing the pre-bound
campaigns as the "before" column. Deliverables on
`claude/ms-eff-table-after-bound` (`57470ba3b`).

## eps_m

| row | before | after |
|---|---|---|
| `PA`/`sequential` | 3.50 | **1.40** |
| `PA`/`sequential_global_retry` | 13.34 | **5.29** |
| `PA (no jac.)`/`sequential` | 1.10 | **1.00** |
| `PA (no jac.)`/`sequential_global_retry` | 3.88 | **3.53** |

Exactly four rows can move: `draw_mass` is PA-only and `upfront` excludes
`joint` and `sequential_with_mass`. **Every control is an exact null** - `PA/joint`,
`PA/sequential_with_mass`, all three `madspin` rows and all three `onshell` rows
are identical in every column *including N_dec to the last digit*. At 200 000
events `madspin/sequential` reads 6.32 before and after. The "offshell keeps the
global bound" decision shows up as a bit-level identity, not an approximate one.

## The 2.5x understates it

The global bound is not reproducible across event counts: same process, same
seed, the probe gives `maxwgts[0] = 9.927` at 200 k against a much tighter value
at 100 k, so before-code `PA/sequential` is **eps_m 10.38 at 200 k but 3.50 at
100 k**. The per-event bound gives **1.40 at both**. So the real reduction is
7.4x at 200 k, and the headline benefit is reproducibility as much as cost.

## Lineshape: no bias

Combined over both resonances, 138 dof, against the replica noise floor measured
on `madspin_density` (139.2 and 135.7):

| cell | after/before | after/weighted | before/weighted |
|---|---|---|---|
| `PA`/sequential | 156.9 | **125.7** | 173.6 |
| `PA`/seq. global retry | 121.3 | 128.9 | 129.6 |
| `PA no-jac`/sequential | 78.3 | 137.3 | 132.1 |
| `PA no-jac`/seq. global | 90.2 | 132.5 | 126.6 |
| every control | **0.0** | - | - |

The largest mover is `PA/sequential`, +12.0 +- 7.1 MeV combined, and its two
resonances move the *same* way (+5.9 +- 10.0 and +18.1 +- 10.1) - the pattern
that would matter, so it was not left there. Combined it is 1.7 sigma, and
against `decay_output = weighted` the after sample sits at 125.7/138 where the
before sits at 173.6/138: the mover ends up **closer** to the unbounded target
than it started. Control cells span 128.6-153.8 against that same reference and
scatter by +-9 MeV in the mean, so +-12 MeV is inside demonstrated noise.

Threshold slices (beta < 0.5, 20 % of the sample): movers at 46.1-92.1 on 69 dof,
both resonances. Nothing there.

Overweight counters at 200 000 events: `PA/seq_global_retry` 3 -> **0**, the other
three movers 0 -> 0, all on 200000/200000 per-event bounds. `madspin` keeps 2 and
2 by design. No "shift plus non-zero counter" signature anywhere.

## Two notes

The `+-0.04` uncertainty in the table caption was violated: under a
geometric-count model the two largest entries are 0.0442 and 0.0406. The caption
now names the model and quotes it as a **floor** rather than a bound, since
event-to-event dispersion adds `2 Var(eps_e)` and `sequential_global_retry`
couples its stages.

`onshell` + `sequential_global_retry` is now in the table with a **dash** in
eps_m: the 3.53 that log line prints there is chain restarts with no mass in them,
identical to the slot-0 draw count, so printing it under a column defined as
virtuality sets would be a wrong number rather than a rounded one.

PR #379, opened 2026-08-20
MadSpin: the reported cross-section now carries the Breit-Wigner truncation

oliviermattelaer:
MadSpin samples each resonance's virtuality only within `BW_cut` widths of the
pole but normalised with the **full** width, so the reported cross-section came
out identical for `BW_cut = 15`, `BW_cut = 1`, or any other value. It now carries
the truncation.

## The factor

`bw_retained_fraction(M, Gamma, N)` evaluates the exact integral of the density
the sampler actually draws from, in `m^2` space:

```
f = [ atan((M^2 - m_min^2)/(M.Gamma)) + atan((m_max^2 - M^2)/(M.Gamma)) ] / pi
```

Both samplers draw `m^2` flat in `R = atan((m^2 - M^2)/(M.Gamma))`, whose full
range is `pi`, so this is the sampler's own normalisation rather than an
approximation of it. For a top at `N = 15` it gives `0.978698`; the linearised
`(2/pi) atan(2N)` form gives `0.97879` and is wrong in the 4th digit of the
headline check below.

Applied per sampled resonance. `onshell`, `onshell_v1` and `none` sample no
virtuality and are not corrected; 2 -> 1 production is guarded out (that guard is
load-bearing - at 2 -> 2 the W factor would have moved the `p p > w+` acceptance
cross-section 100521.5 by 2212 pb against an `error` of 800).

The nested-resonance treatment differs by path, deliberately: the density modes
correct the top only (`0.95785` for `t > w+ b, w+ > e+ ve`), because they divide
MG5-measured partial widths that already carry the nested truncation, while
`madspin_v1` corrects top and W (`0.91614`) because it uses the param card's
untruncated chain BR. Verified empirically on all three configurations.

## Headline: sigma now depends on BW_cut

`p p > t t~ j`, 30 000 events, `spinmode PA`:

| | reported sigma |
|---|---|
| `BW_cut = 15` | 645.741 pb |
| `BW_cut = 1` | 334.910 pb |
| measured ratio | **1.928103** |
| predicted `f(15)^2/f(1)^2` | **1.928102** |

Seven digits. Dividing the factor back out recovers 674.157 / 674.162 against the
production sample's 674.1565 pb. `onshell` reports 674.1565 pb at both cuts,
unchanged.

## Against an independent truth

Truth generated as `p p > t t~ j, t > w+ b, t~ > w- b~` at matched `bwcutoff`,
where MG5 *rejects* out-of-window points so the truncation is in its integrated
cross-section.

| | truth/MadSpin before | after |
|---|---|---|
| `BW_cut = 15` | 0.9688 | **1.0114 +- 0.0024** |
| `BW_cut = 1` | 0.4971 | **1.0006 +- 0.0027** |

The second `BW_cut` settles a question the earlier analysis left open. It had
bounded a residual to +0.4%..+1.0% and attributed it to the decay numerator
`m.Gamma(m)/(m_t.Gamma_t)` running 0.52 -> 1.71 across the +-15 Gamma window while
the factor holds it flat. That explanation *predicts* the residual collapses as
the window narrows, and it does: **+1.1% at N=15, +0.06% at N=1** - the latter
indistinguishable from zero in the run where the correction removes half the
rate. So it is a window-width effect, not an offset to live with. Folded against
the earlier 5M-event samples the sharpest estimate is **+0.90% +- 0.05%**.

## Tests

`tests/test_manager.py test_madspin -t0`: **444 tests, OK**.

A full sweep of every cross-section / BR / event-weight assertion in `tests/`
found exactly **one** affected number, and it did not fail:
`test_cmd_madevent.py::test_complex_mass_scheme`, whose post-`decay_events`
target moves 440.779 -> 431.39 (-2.13%), exactly
`bw_retained_fraction(173.0, 1.491257, 15) = 0.9786983` to seven digits. Its
tolerance is +-4.3% on a 100-event run and swallowed the shift; updated anyway,
since a tolerance wide enough to hide a systematic is not a check on it. The two
targets were only ever equal because BR(t -> W b) = 1 to seven digits.

## Not covered

`test_madspin_factory` reviewed by reading, not run. Spinmodes `full`, `none`,
`onshell_v1` not exercised end to end. `decay_output = weighted`, mixed-pdg BR
equalisation and fixed-order composition are argued in the doc, not re-measured.

One latent inconsistency recorded for the future: `bw_retained_fraction` floors
`m_min` at 0.0, `_mass_window` applies no floor, and `generate_momenta` floors at
0.5. All three agree exactly whenever `N.Gamma < M`, which covers every physical
case - it would only matter for a resonance with `Gamma > M/15`.

## Summary by Sourcery

Make MadSpin report cross-sections that include the Breit–Wigner truncation imposed by its sampled resonance-mass windows.

New Features:
- Scale reported MadSpin cross-sections and branching ratios by the Breit–Wigner fraction retained by the configured virtuality window.
- Apply truncation corrections consistently across density and v1 decay paths while excluding modes and production topologies that do not sample virtualities.
- Expose Breit–Wigner truncation metadata and untruncated comparison values in MadSpin parallel-test results.

Bug Fixes:
- Ensure reported cross-sections now vary with BW_cut and remain consistent with the portion of the resonance distribution actually generated.

Enhancements:
- Use an exact m²-space Breit–Wigner retained fraction and account for nested resonances according to each sampling and branching-ratio convention.
- Update cross-section consistency checks to compare truncation-corrected values and validate mode-specific truncation behavior.

Documentation:
- Document the Breit–Wigner truncation behavior, validation results, affected spinmodes, and remaining limitations.

Tests:
- Add unit coverage for the retained-fraction calculation, BW_cut handling, spinmode selection, broad resonances, and v1 nested-chain corrections.
- Update acceptance expectations and parallel MadSpin comparisons for truncation-aware cross-sections.


oliviermattelaer:
One caveat worth weighing before merging, from the validation pass.

**This change is user-visible but silent in CI.** The only existing test that
touches the moved number - `test_complex_mass_scheme` - passed both before and
after, because its tolerance is `4.err` = +-4.3 % on a 100-event run and the shift
is -2.13 %. So the reach is genuinely small (1 expectation changed, 0 tests
broken), but the flip side is that **the suite would not have caught a wrong
truncation factor** either.

The unit tests added here cover `bw_retained_fraction` itself, and
`madspin_comparator.py` gained `assert_bw_truncation_matches_spinmode`, but
nothing in the default suite asserts end to end that the reported cross-section
actually scales with `BW_cut`. The headline check in the PR body -
`sigma(15)/sigma(1) = 1.928103` against a predicted `1.928102` - was run by hand.

That is cheap to close: a test that runs a small sample at two `BW_cut` values and
asserts the ratio against `bw_retained_fraction` would pin the factor, the
per-resonance product and the mode gating in one assertion. Happy to add it if
wanted - say the word and it goes on this branch.

PR #390, opened 2026-08-27
MadSpin: let keep_weight_for_polarization_vector carry the axial state (axial 4/4)

oliviermattelaer:
**Milestone 4 of 4**, the last of the axial feature. Stacked on #389. Lets
`keep_weight_for_polarization_vector` carry the axial state, replacing the
refusal #389 put in.

## The decision: `A` is a genuine extra weight

#389 refused the combination because "no projection onto a physical polarisation
can pick up the axial direction, so the set stops adding to one". True in its
first half, and answered by its second: the axial direction **is** a basis
direction, and naming it closes the list.

The nominal contraction under `consider_axial` is
`W = Sum_{i,j in B4} c_i c_j rho_P(i,j) rho_D(i,j)` over `B4 = {-1,0,+1,A}`, and
every entry of this option is that sum with the rows masked to one block. For a
single state that is `c_lambda^2 rho_P rho_D`, which for `lambda = A` is
non-negative and is computed by the code path the other entries already use —
#389 folds `c_i c_j` into the decay matrix, so nothing is renormalised and nothing
is invented.

**The sum rule is repaired, not weakened.** Its two conditions — the list
partitions each slot's basis, and no cross-block interference — are unchanged.
`consider_axial` widens the basis, so `[0,+,-]` *stops* being a partition and
`[0,+,-,A]` becomes one.

**Other entries do not move**: 600/600 `<wgt>` entries bit-identical between
`[0,+,-,A]` and `[0,+,-]` on the same seed.

## The caveat, recorded rather than glossed

The axial direction reaches the contraction **twice** — through
`2 Re Sum_lambda c_A rho_P(lambda,A) rho_D(lambda,A)`, which is `O(c_A)`, and
through `c_A^2 rho_P(A,A) rho_D(A,A)`. `w_A` is only the second. So it is by
construction the *smaller* piece and is **not** a measure of "how much
`consider_axial` matters" — it is the axial member of this option's diagonal
decomposition, no more. In the code and the notes.

A consumer must also not read it as a cross-section: every other entry is one
(rerun with a `{0}` brace), and there is no `{A}` sample — which is exactly why
#389 keeps the axial row out of every trace. That goes into the banner's
`<weightgroup>` as XML comments and into the weight's own description, and the
column is spelled `axial`, not `4`.

## Numbers

`u d~ > w+ h`, `decay w+ > ta+ vt`, spinmode madspin, `consider_axial True`, 200
events, summed over the sample / nominal:

| | `[0,+,-,A]` | `[0,+,-]` | `consider_axial` off |
|---|---|---|---|
| `w+(0)` | 56.2906 % | 56.2906 % | 56.2921 % |
| `w+(+)` | 25.2833 % | 25.2833 % | 25.2833 % |
| `w+(-)` | 23.9154 % | 23.9154 % | 23.9174 % |
| `w+(axial)` | 0.00026 % | — | — |
| **sum** | **105.4896 %** | **105.4894 %** | **105.4928 %** |

The sums are not 100 %, and the axial direction is not why: that 5.5 % is the
`(-1,+1)` interference of a 1->2 decay and is already there with `consider_axial`
off. `w_A >= 0` on all 200 events, and **binned in `|c_A|` it follows `c_A^2` over
three decades — fitted log-log slope 2.098 against the 2 the formula predicts.**
That is the closure test on "no invented normalisation".

## Also

`_POL_TOKENS` is now *derived* from `POLARIZATION_ALIASES` minus the axial entry
(verified identical to the old literal) instead of being a second hand-written
copy of the same vocabulary. `A` with `consider_axial` off is refused at launch;
a vector list omitting `A` with the option on warns. #389's other refusals stand.

## Tests and bit-identity

`test_madspin -t0` **509 OK** (498 on base, +11). `-p U test_cmd` 24 OK. Full
`-p U` 1295, only the pre-existing environmental `test_DensityMatrixObservables22`.
Against the base with `consider_axial` off, decayed LHE byte-identical on both a
plain run (207472 bytes) and one with `[0,+,-]` (237792 bytes).

## Found, not fixed

`BW_cut = 200` dies in an unbounded retry recursion in
`lhe_parser.Event.reshuffle_production` (`_allow_retry` is never decremented).
Predates this branch and is unrelated to it; filed separately.

## Summary by Sourcery

Allow polarization weights to include the axial vector direction under `consider_axial`, while preserving compatibility checks and clearly documenting its interpretation.

New Features:
- Support the axial polarization label `A` in polarization-weight requests when `consider_axial` is enabled.
- Emit axial polarization weights with descriptions clarifying that they represent propagator spin-correlation contributions rather than observable cross-sections.

Bug Fixes:
- Restore complete polarization-weight partitioning for widened four-state vector bases by allowing lists that include the axial direction.

Enhancements:
- Warn when axial polarization weights are requested without listing `A`, and reject `A` when axial mode is disabled.
- Derive the pure-interference polarization vocabulary from the shared polarization alias definitions while excluding the axial label.
- Extend polarization layout, combination handling, and banner generation to recognize axial vector basis entries.

Documentation:
- Document the axial polarization label, its sum-rule behavior, and its non-cross-section interpretation in MadSpin options and event-file weight metadata.

Tests:
- Add coverage for axial label parsing, compatibility checks, basis filtering, sum rules, non-negativity, multi-slot combinations, and banner descriptions.

### DIFF
--- a/MadSpin/interface_madspin.py
+++ b/MadSpin/interface_madspin.py
@@ -58,9 +58,14 @@ cmd_logger = logging.getLogger('cmdprint2') # -> print
 # Polarisation labels accepted by keep_weight_for_polarization_*
 # ---------------------------------------------------------------------------
 # Same spelling and same meaning as MG5's polarisation braces:
-#   {L} -> [-1], {R}/{+} -> [1], {T} -> [-1,1], {0} -> [0]
+#   {L} -> [-1], {R}/{+} -> [1], {T} -> [-1,1], {0} -> [0], {A} -> [4]
 # Maps the (case-insensitive, brace-tolerant) user spelling onto
 # (canonical label, helicity values).
+#
+# '{A}' is the AXIAL direction eps_A = k/sqrt(k.k), HELAS' nhel = 4. It only
+# exists in a basis 'set consider_axial True' has widened -- every other slot
+# drops it the way a fermion drops '0' -- and a card that names it without that
+# option is refused in _check_axial_compatibility rather than silently ignored.
 POLARIZATION_ALIASES = {
     '0': ('0', (0,)),
     '+': ('+', (1,)),
@@ -68,13 +73,21 @@ POLARIZATION_ALIASES = {
     '-': ('-', (-1,)),
     'l': ('-', (-1,)),
     't': ('T', (-1, 1)),
+    'a': ('A', (madspin.DensityMatrix.AXIAL_HELICITY,)),
 }
+
+#: canonical labels whose helicity values are the *physical* polarisations, i.e.
+#: everything except the axial direction. What the pure-interference vocabulary
+#: is allowed to speak (see MadSpinInterface._POL_TOKENS).
+PHYSICAL_POLARIZATION_ALIASES = {
+    key: value for key, value in POLARIZATION_ALIASES.items()
+    if value[0] != 'A'}
 
 
 def parse_polarization_label(label):
     """(canonical label, helicity values) for one
     keep_weight_for_polarization_vector/_fermion entry, or None if it is not
-    one of 0/+/-/T (L and R aliasing - and +)."""
+    one of 0/+/-/T/A (L and R aliasing - and +)."""
     key = str(label).strip().lower()
     if key.startswith('{') and key.endswith('}'):
         key = key[1:-1].strip()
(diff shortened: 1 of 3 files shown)


MadSpin/interface_madspin.py
```diff
@@ -7560,16 +7611,76 @@ def _check_axial_compatibility(self):
...
+            if vector and 'A' not in vector:
+                logger.warning(
+                    "MadSpin: consider_axial is on and "
+                    "keep_weight_for_polarization_vector = %s does not name "
+                    "'A', so it no longer partitions a massive vector's spin "
+                    "basis: the axial block of the convolution belongs to none "
+                    "of these weights and their sum is short by it. Add 'A' to "
+                    "the list to close the partition. (That is condition (a) "
+                    "of the sum rule; condition (b), the interference terms, "
+                    "is unaffected either way.)", vector)
```
oliviermattelaer:
```suggestion
```

PR #404, opened 2026-09-07
Fix truncation of long run_card lines in the NLO LHE banner (bug #2046152)

oliviermattelaer:
Fixes the remaining part of [bug #2046152](https://bugs.launchpad.net/bugs/2046152) ("systematics module warning/crash at NLO").

## The bug

The NLO banner code reads LHE/card lines into **fixed-length Fortran buffers** and writes them straight back out. Any line longer than the buffer is silently cut in the middle of the value. For a long `systematics_arguments` entry -- the standard ATLAS string is 565 characters -- the surviving fragment is no longer a valid `value = name` pair, so re-reading the file gives

```
Found unexpected entry in run_card: "nnpdf30_nlo_as_0118_hessian,pdf4lhc" with value "['--weight_info=...','--pdf"
```

or, depending on where the cut lands, silently drops the entry (it comes back as `['']`).

Two places are affected:

| routine | buffer | status |
|---|---|---|
| `copy_header` | 200 chars | **live** -- used by the `split_events`, `merge_events` and `select_events` utilities |
| `write_lhef_header_banner` | 250 chars | dead code since 3.7.1 (event collection moved to `collect_events.py`), but still live in the LTS series |

The originally reported symptom (the banner of `Events/<run>/events.lhe.gz` itself) is already gone on this branch, because `write_lhef_header_banner` is no longer called. `copy_header` still truncates today.

## Reproducer

`p p > e+ ve [QCD]`, with Zach's 565-character `systematics_arguments` in the run_card.

* **v3.7.0** (last release using the fortran banner writer): the `systematics_arguments` line in `events.lhe` is cut at exactly 250 characters, and `Banner.charge_card('run_card')` returns `['']`.
* **3.8.0, before this PR**: `events.lhe` is fine (565 chars), but running `Utilities/split_events` on it cuts the line to 200 characters, and the split files again parse back as `['']`.
* **3.8.0, with this PR**: `split_events` output keeps the full 565 characters and parses back identically to the input. Apart from the intended `nevents` replacement, the copied header is byte-identical to the input header.

## The fix

Instead of picking a larger magic number, records are read in chunks with non-advancing I/O (`advance='no'` + `size=` + `eor=`) and copied over in full, so **lines of arbitrary length are preserved** and no buffer size can ever be exceeded again.

New helper routines in `handling_lhe_events.f`: `read_record_head`, `copy_rest_of_record`, `skip_rest_of_record`, `write_record`, `copy_open_file`. `write_lhef_header_banner` is converted to them too, so the patch is directly backportable to the LTS branch where that routine is still used.

Note: because the copy is now exact, banner lines keep the trailing blanks present in the source file (`copy_header` used to `trim` them -- that was mostly there to undo the blank padding of the old fortran banner writer). The iseed/nevents substitutions are only attempted on records that fit in the buffer, which cannot change behaviour since those entries are short.

Verified: `gfortran -Wall` clean (no new warnings), `split_events` / `merge_events` / `select_events` build and run, and a full `p p > e+ ve [QCD]` aMC@NLO run is unaffected.

## Summary by Sourcery

Prevent truncation of long records when generating and copying NLO LHE banners.

Bug Fixes:
- Preserve arbitrarily long LHE banner and card records so entries such as long systematics arguments remain valid when copied and reread.

Enhancements:
- Consolidate banner and file-copy handling around record-aware routines that retain exact source content while still updating supported short metadata entries.

**The buffers stay at 250 characters.** Their length is no longer a limit on the record length — it only sets how much of a record is read at a time, and how much of it is inspected for the (short) `iseed`, `nevents` and `ickkw` entries. A narrower window is if anything safer there. Verified to give byte-identical output to a 1000-character version.

### DIFF
--- a/Template/NLO/SubProcesses/handling_lhe_events.f
+++ b/Template/NLO/SubProcesses/handling_lhe_events.f
@@ -149,6 +149,8 @@ c     other parameter
       character*10 MonteCarlo
       character*100 path
       character*250 buffer,buffer_lc,buffer2
+      integer nread
+      logical complete,eof
       integer event_id
       common /c_event_id/ event_id
       integer n_orderstags,oo,tag
@@ -165,20 +167,14 @@ c
       write(ifile,'(a)') '  <MG5ProcCard>'
       open (unit=92,file=path(1:index(path," ")-1)//'proc_card_mg5.dat'
      &     ,err=99)
-      do
-         read(92,'(a)',err=89,end=89) buffer
-         write(ifile,'(a)') buffer
-      enddo
- 89   close(92)
+      call copy_open_file(92,ifile)
+      close(92)
       write(ifile,'(a)') '  </MG5ProcCard>'
       write(ifile,'(a)') '  <slha>'
       open (unit=92,file=path(1:index(path," ")-1)//'param_card.dat'
      &     ,err=98)
-      do
-         read(92,'(a)',err=88,end=88) buffer
-         write(ifile,'(a)') buffer
-      enddo
- 88   close(92)
+      call copy_open_file(92,ifile)
+      close(92)
       write(ifile,'(a)') '  </slha>'
       write(ifile,'(a)') '  <MGRunCard>'
 c     import the parameter from the run_card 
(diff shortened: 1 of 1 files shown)


oliviermattelaer:
done in 3.8.0 via 3.5.17 merge

PR #406, opened 2026-09-08
[3.5.x] Do not apply the beam polarization twice (bug #2148181)

oliviermattelaer:
Fixes https://bugs.launchpad.net/mg5amcnlo/+bug/2148181 (reported by Richard Ruiz).

## The bug

`ve{L} d > e- u` gives exactly twice the cross section of `ve d > e- u`.

Reproduced on `3.5.17` (`eb1eecb03`):

| process | run_card `polbeam1` | cross section |
|---|---|---|
| `ve d > e- u`    | -100 (auto) | 93.22 +- 0.78 pb |
| `ve{L} d > e- u` | -100 (auto) | 186.4 +- 1.56 pb |

## Cause

Two independent mechanisms select the left-handed neutrino, and both were applied:

* the process definition `ve{L}` reduces the spin-averaging denominator, `get_denominator_factor()` in `madgraph/core/helas_objects.py` uses `len(leg.get('polarization'))` instead of the number of helicity states. `IDEN` goes from 12 to 6 and `NCOMB` from 16 to 8;
* `RunCard.create_default_for_process()` automatically sets `polbeam1 = -100` for a neutrino beam, and the generated `matrix<i>.f` reweights each helicity by `POL(JJ)`, which is `+-2` for a 100% polarized beam.

Hence the extra factor 2.

## Fix

`POL(JJ)` is really `2 x (fraction of the beam in that helicity)`; the `2` is there to undo the `1/2` of the initial-state spin average. When the process definition already restricts that beam to a single helicity state, there is no `1/2` left to undo, so the reweighting has to be rescaled by `NB_SPIN_STATE_IN(JJ)/2`:

```fortran
T=T*ABS(POL(JJ))*NB_SPIN_STATE_IN(JJ)/2d0
```

This normalization was already present in the helicity-recycling template (`matrix_madevent_group_v4_hel.inc`, introduced in 05946d470) but had never been propagated to the other madevent templates. This PR applies it to:

* the grouped and non-grouped madevent output (`beam_polarization` snippet in `export_v4.py`, both the two-beam and the single-beam variant);
* `matrix_loop_induced_madevent.inc` and `matrix_loop_induced_madevent_group.inc`, which hardcode the same block. The non-grouped loop-induced template now declares the common block itself, filled from the new `nb_spin_state1/2` entries of the loop-induced `replace_dict`.

`matrix_madevent_v4.inc` renames its common-block variable `nb_spin_state` to `nb_spin_state_in` (the name the grouped template already uses), so that the shared code snippet compiles in both.

Rather than forbidding the combination, this keeps `polbeam` meaningful and composable: for a partially polarized beam the weight becomes the actual helicity fraction instead of twice it.

## Validation

Full `generate_events` runs, same run_card in each pair:

| process | `polbeam1` | before | after |
|---|---|---|---|
| `ve d > e- u`      | -100 | 93.22 +- 0.78 pb | 93.22 +- 0.78 pb |
| `ve{L} d > e- u`   | -100 | **186.4 +- 1.56 pb** | **93.22 +- 0.78 pb** |
| `e- e+ > mu+ mu-`      | -100 | | 0.1145 +- 0.0002 pb |
| `e-{L} e+ > mu+ mu-`   | -100 | | 0.1145 +- 0.0002 pb |
| `e- e+ > mu+ mu-` (no group)    | -100 | | 0.1139 +- 0.0002 pb |
| `e-{L} e+ > mu+ mu-` (no group) | -100 | | 0.1139 +- 0.0002 pb |

Unpolarized processes are untouched: `NB_SPIN_STATE_IN = 2` gives a factor 1.

Also checked:

* `generate g g > h [noborn=QCD]` output, grouped and non-grouped, `make matrix1.o` / `make matrix.o` compile clean;
* subprocess grouping keeps polarized and unpolarized processes in separate `P*` directories (`generate ve d > e- u` + `add process ve{L} s > e- c` gives `P1` with `NB_SPIN_STATE /2,2/` and `P2` with `/1,2/`), so the per-directory common block is always consistent;
* the two `IOExportV4IOTest` madevent reference files are updated;
* `tests/test_manager.py -p tests/unit_tests`: 665 tests, the only failures are the pre-existing python 3.14 signed-zero `JAMP` mismatches and the unrelated `test_usermod` import error, all present before this change.

## Summary by Sourcery

Prevent beam polarization from being double-counted when process definitions restrict incoming helicity states.

Bug Fixes:
- Correct beam-polarization normalization so explicitly polarized process definitions no longer receive the beam polarization factor twice.
- Preserve correct cross sections for unpolarized and partially polarized beams across grouped, non-grouped, and loop-induced generated processes.

Enhancements:
- Propagate the number of retained initial-state helicity states through loop-induced exports and standardize the shared helicity-state common-block name.

Tests:
- Update madevent reference outputs to reflect the corrected polarization normalization.

### DIFF
--- a/madgraph/iolibs/export_v4.py
+++ b/madgraph/iolibs/export_v4.py
@@ -4633,10 +4633,12 @@ class ProcessExporterFortranME(ProcessExporterFortran):
         if self.beam_polarization == [True, True]:
             replace_dict['beam_polarization'] = """
                          DO JJ=1,nincoming
+c NB_SPIN_STATE_IN/2 avoids a double counting
+c of an explicit polarisation in the process
                IF(POL(JJ).NE.1d0.AND.NHEL(JJ,I).EQ.INT(SIGN(1d0,POL(JJ)))) THEN
-                 T=T*ABS(POL(JJ))
+                 T=T*ABS(POL(JJ))*NB_SPIN_STATE_IN(JJ)/2d0
                ELSE IF(POL(JJ).NE.1d0)THEN
-                 T=T*(2d0-ABS(POL(JJ)))
+                 T=T*(2d0-ABS(POL(JJ)))*NB_SPIN_STATE_IN(JJ)/2d0
                ENDIF
              ENDDO
             """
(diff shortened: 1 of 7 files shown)


richardphysics (approved):
proposed updates look good!

PR #422, opened 2026-09-16
UPC with neutron tagging

lsimon125:
Upgrade of gamma-UPC to version 1.8.0 with additional neutron tagging option (by Nicolas Crépet, David d'Enterria, Luca Maxia, Hua-Sheng Shao)

## Summary by Sourcery

Upgrade gamma-UPC to v1.8 and integrate configurable forward neutron tagging across UPC run-card, PDF, build, and validation workflows.

New Features:
- Add configurable forward neutron-tagging modes for UPC photon fluxes, including 0n, 1n, 2n, 3n, 4n, and Xn combinations with central, upper, and lower uncertainty variants.
- Provide photoabsorption and neutron-emission cross-section models for Pb-208, Au-197, U-238, and O-16 to support neutron-tagged calculations.

Bug Fixes:
- Update gamma-UPC integration and interpolation behavior for the version 1.8 implementation.

Enhancements:
- Upgrade the embedded gamma-UPC library to version 1.8 and expose neutron-tagging uncertainty control through run-card PDF labels.
- Validate supported ion species and neutron-tagging combinations and reject unsupported configurations with clear errors.
- Extend photon-PDF generation and systematics handling to recognize the expanded gamma-UPC label set.

Build:
- Update LO and NLO build rules to link gamma-UPC for all supported neutron-tagging PDF labels.

Chores:
- Improve diagnostics emitted when the gamma-UPC dummy implementation is linked.


oliviermattelaer:
My first feedback would be:
 1. that some long list need to be shorten and some part of the code as well
 2. That nothing is testing such feature (meaning no way to ensure that they will work when merging within MG7)
 
Asking suggestion to Claude for those two points (and he reports some potential issue that I report here but not evaluated).





## A. Simplification requests

### A1. The 300-label triple loop is duplicated 7 times

The exact same block

```python
upclabels=['edff','chff','edpf','edmf','chpf','chmf']
for nerr in ['f','p','m']:
    for y in ['i','x','0','1','2','3','4']:
        for z in ['i','x','0','1','2','3','4']:
            upclabels.append('ed'+nerr+y+'n'+z+'n')
            upclabels.append('ch'+nerr+y+'n'+z+'n')
```

appears in:

| file | location |
|---|---|
| `madgraph/various/banner.py` | `PDLabelBlock.check_validity` (4227) |
| `madgraph/various/banner.py` | `RunCardLO.default_setup` valid_pdf (4509) |
| `madgraph/various/banner.py` | `RunCardLO.check_validity`, **inside the `for i in [1,2]` loop** (4853) — rebuilt twice per call |
| `madgraph/various/banner.py` | `RunCardLO.check_validity` fixed_fac_scale warning, **inside a loop and an `if`** (4938) |
| `madgraph/various/banner.py` | `RunCardNLO.default_setup` valid_pdf (5932) |
| `madgraph/interface/madevent_interface.py` | compile/make_opts stage (6576) |
| `madgraph/interface/amcatnlo_run_interface.py` | make_opts stage (5520) |
| `madgraph/various/systematics.py` | `nosystlabels` (172) |

Proposal: one module-level definition in `banner.py`, reused everywhere:

```python
# banner.py, module level (next to the other constants)
UPC_LABELS = ['%s%s%s' % (ff, err, tag)
              for ff in ('ed', 'ch')
              for err in ('f', 'p', 'm')
              for tag in ['f'] + ['%sn%sn' % (y, z) for y in 'ix01234' for z in 'ix01234']]

def is_upc_label(label):
    return label in UPC_LABELS      # or: re.fullmatch(r'(ed|ch)[fpm](f|[ix0-4]n[ix0-4]n)', label)
```

(I verified the regex/comprehension reproduce exactly the 300 labels of the loops.)
Then `banner.UPC_LABELS` / `banner.is_upc_label()` in the two interfaces and in
`systematics.py`, which already imports banner.

Bonus: the `'edff'` still hard-coded at `banner.py:4002` (get_pdf_id → 0) and `:5112`
(default pdlabel for a a → `edff`) are fine as-is, but the comment at 4004 is misleading:
the `except: return 0` catches *any* unknown label, not only UPC ones.

### A2. The nucleus/neutron-tagging check is written 4 times (2 beams × 2 files)

`PDLabelBlock.check_validity` (banner.py 4272-4310) and
`MadEventCmd.compile_...` (madevent_interface.py 6588-6632) contain the same
beam-1 / beam-2 blocks, each a copy of the other with `1`→`2`. Also:

* In banner.py the outer condition `card['pdlabel1'] in upclabels` is immediately
  re-tested in the inner `if card['pdlabel1'] in upclabels:` — the `else: beam1_n='in'`
  branch is dead code.
* In madevent_interface the `elif self.run_card['pdlabel2'] in upclabels` branch is
  dead too, because `PDLabelBlock.check_validity` (run on every card write) already
  forces `pdlabel2 = pdlabel1` when one of them is a UPC label.
* madevent_interface raises a bare `Exception`; banner raises `InvalidRunCard`.
* The madevent_interface copy is redundant with the banner copy — banner's
  `check_validity` always runs before `compile`. Suggest deleting the interface copy
  entirely (or, if the authors want a belt-and-braces check at compile time, call a
  shared helper).

Proposed single helper (banner.py):

```python
# (Z, N) -> highest tagged multiplicity implemented; None = any of xn,0n..4n
UPC_NEUTRON_TAGGING_NUCLEI = {(82, 126): 4, (79, 118): 4, (8, 8): 2}

def check_upc_neutron_tagging(card):
    """raise InvalidRunCard if pdlabel asks for a neutron tag not implemented for the beam ion."""
    label = card['pdlabel1'] if card['pdlabel1'] in UPC_LABELS else card['pdlabel2']
    if not (card['lpp1'] == 2 == card['lpp2'] and label in UPC_LABELS and label[3] != 'f'):
        return
    for i, tag in ((1, label[3:5]), (2, label[5:7])):
        if tag == 'in':
            continue
        Z, N = card['nb_proton%i' % i], card['nb_neutron%i' % i]
        if (Z, N) == (1, 0):
            continue                     # proton: tag ignored by gamma-UPC
        nmax = UPC_NEUTRON_TAGGING_NUCLEI.get((Z, N))
        if nmax is None or (tag[0].isdigit() and int(tag[0]) > nmax):
            raise InvalidRunCard("%s: neutron tagging '%s' not implemented for beam %i (Z=%s, N=%s)"
                                 % (label, tag, i, Z, N))
```

This replaces ~80 lines with ~15 and gives a message that says *which* beam and *why*
(the current message "check your two beam particle species" does not).

### A3. Makefile filter lists (300 literal words, three times)

`Template/LO/Source/PDF/makefile`, `Template/NLO/Source/makefile`,
`madgraph/iolibs/template_files/madevent_makefile_source` each got a
~2.4 kB single-line `$(filter edff chff edpf ... chm4n4n, ...)`.

GNU make's `filter` accepts `%` wildcards, and no other pdlabel starts with `ed` or
`ch` (checked: lhapdf, emela, cteq*, ct14*, nn23*, iww, eva, none, mixed, dressed,
and the lep_densities directory names). So the three lines can become

```make
ifneq (,$(filter ed% ch%, $(pdlabel1) $(pdlabel2)))
```

I verified with a scratch makefile that `$(filter ed% ch%, edf1n2n chmf nn23lo1 cteq6l1 eva)`
returns `edf1n2n chmf`.

**Actual bug found while diffing the lists**: the NLO makefile list is missing
`edpf edmf chpf chmf` (49 instead of 50 entries per prefix). With the current PR,
an NLO run with `pdlabel = edpf` will not link libgammaUPC and will hit the dummy
`STOP 'gamma-UPC 1'` at run time. The wildcard fix removes this class of error.

### A4. Fortran parser in `photonpdfsquare.f`

The `ed` and `ch` branches are ~60 lines each and identical except for
`USE_CHARGEFORMFACTOR4PHOTON` and the STOP codes. A small internal function
`neutron_tag_code(c)` (character → -2..4, STOP on unknown) would collapse the two
7-way `elseif` ladders ×2 branches into 2 calls each. Not blocking, but the file is
now 320 lines for what is a 7-character grammar.

Also: `STOP -1` / `STOP -2` — negative stop codes are accepted by gfortran but the
process exit status becomes 255/254; the other branches use small positive codes.
Cosmetic, but they are also the same numeric codes (2,3) reused for different errors
(STOP 2 "unknown pdlabel" and STOP 3 "bad nb_hadron" / "bad 5th-7th char").

### A5. Inconsistent prefix tests between LO and NLO exporters

* `export_v4.py` now emits `PDLABEL(1:3).EQ.'edf'.OR.'chf'.OR.'edp'.OR.'edm'.OR.'chp'.OR.'chm'`
* `export_fks.py` emits `PDLABEL(1:2).EQ.'ed'.OR.'ch'`

Both are correct (no other label starts with ed/ch) but they should be the same; the
`(1:2)` form is shorter and matches `photonpdfsquare.f`.

## C. Does this need tests? Yes.

No test is added, and nothing existing exercises `edff`/`chff` at all (grep over
`tests/` finds only the two orphan input cards). The PR changes card validation logic,
makefile generation, Fortran label parsing, and adds a nucleus×tag compatibility
matrix — all of which are cheap to unit-test and are exactly the kind of thing that
regresses silently (the NLO makefile omission above is a demonstration).

### C1. Unit tests (fast, no compilation)

`tests/unit_tests/various/test_banner.py` (extend `TestRunCard` / `test_pdlabel_block`):

1. **`test_upc_label_set`** — `banner.UPC_LABELS` has exactly 300 entries, all match
   `(ed|ch)[fpm](f|[ix0-4]n[ix0-4]n)`, all ≤ 7 chars (Fortran `character*7`), and
   every one is in `RunCardLO()['pdlabel'].allowed` and `RunCardNLO()['pdlabel'].allowed`.
2. **`test_upc_pdlabel_block_propagation`** — LO card, `lpp1=lpp2=2`,
   `set('pdlabel1','edf1n2n')` → after `check_validity()` `pdlabel == pdlabel2 == 'edf1n2n'`;
   same via `set('pdlabel', ...)`; same with `chp0nxn`.
3. **`test_upc_pdlabel_lpp_compat`** — `lpp1=1` with `pdlabel1='edfinin'` raises
   `InvalidRunCard`; `lpp1=2` with `pdlabel1='nn23lo1'` is silently changed to `edff`
   (warning path).
4. **`test_upc_neutron_tagging_nuclei`** — parametrised over
   (Z,N,label,expect): Pb208/`edf4n4n` OK, Au197/`chm3n1n` OK, O16/`edf2n2n` OK,
   O16/`edf3n0n` raises, O16/`edf0n4n` raises, proton (1,0)/`edf1n1n` OK (ignored),
   Xe129 (54,75)/`edfxnxn` raises, any nucleus/`edfinin` OK, any nucleus/`edff`/`edpf` OK;
   asymmetric: beam1 Pb + beam2 O with `edf3n2n` OK but `edf2n3n` raises.
   (Decide U238 first, then add it to the table on the chosen side.)
5. **`test_upc_neutron_tagging_nlo`** — same matrix through `RunCardNLO` once B3 is fixed.
6. **`test_get_pdf_id_upc`** — `get_pdf_id` (line ~3990) returns 0 for every UPC label.
7. **`test_upc_run_card_write_read`** — write a card with `pdlabel1='chm4n4n'`, re-read
   it with `banner.RunCard(path)`, check the value survives and `run_card.inc` output
   contains `pdsublabel(1) = 'chm4n4n'` (7 chars, no truncation).

`tests/unit_tests/various/test_systematics.py` (new or in test_misc):

8. **`test_systematics_refuses_upc`** — `Systematics` raises `SystematicsError` for
   pdlabel in (`edff`, `edpf`, `chf1n2n`, `chm4nxn`), passes for `nn23lo1`.

`tests/unit_tests/various/test_make_opts.py` or `tests/unit_tests/iolibs/test_export_v4.py`:

9. **`test_makefile_source_links_gammaupc`** — render
   `madevent_makefile_source` / the two Template makefiles, then run
   `make -n -f <file> pdlabel1=edf1n2n` style dry-runs (or a tiny scratch makefile
   that includes the `ifneq` line) for a sample of labels in each of the six prefix
   classes (`edff, edpf, chmf, edf1n2n, chp0nxn, edm4n4n`) plus negatives
   (`nn23lo1`, `eva`, empty) and assert whether `libgammaUPC` appears in the
   dry-run output. This is the test that would have caught the NLO omission.

`tests/unit_tests/iolibs/test_export_v4.py` / `test_export_fks.py`:

10. **`test_pdf_lines_upc_condition`** — for a `a a > ...` process the generated
    `pdf_lines` contain the PHOTONPDFSQUARE branch and the condition uses the agreed
    prefix test (A5), identically in LO and NLO.

Fortran-level (compile the gammaUPC directory only, ~seconds; fits `unit_tests`):

11. **`test_photonpdfsquare_label_parser`** — small Fortran driver linked against
    `libgammaUPC` that sets `pdlabel` to each of the 300 labels and checks
    `neutron_tagging(1:2)` / `neutron_xsigma` / `USE_CHARGEFORMFACTOR4PHOTON` come
    out as expected, and that `edfq1n`, `edg`, `edf1n2x` stop. (Could be a single
    driver reading labels from stdin, run once.)

### C2. Acceptance tests (compile + run, mark slow)

`tests/acceptance_tests/test_cmd_madevent.py`:

1. **`test_upc_pbpb_edff_baseline`** — `generate a a > mu+ mu-`, output, run_card
   `lpp=2/2`, `nb_proton=82`, `nb_neutron=126`, `ebeam=2510` per nucleon or the
   default, `pdlabel=edff`, `nevents=100`, `nb_core=1`; check the cross section
   against a stored reference within stat error. This is the regression anchor for
   the gamma-UPC 1.7→1.8 upgrade (baseline currently absent).
2. **`test_upc_pbpb_neutron_tagging`** — same process with `edf0n0n`, `edfxnxn`,
   `edf1nin`; assert σ(0n0n) < σ(inclusive), σ(xnxn) < σ(inclusive), and
   σ(0n0n)+σ(0nxn)+σ(xn0n)+σ(xnxn) ≈ σ(inin) to a few percent (probabilities of the
   four exclusive break-up classes must sum to one). That last relation is a
   physics check that costs nothing extra.
3. **`test_upc_neutron_uncertainty_band`** — `edp1n1n` and `edm1n1n` bracket
   `edf1n1n` (σ_m ≤ σ_f ≤ σ_p, or the reverse; assert ordering is monotonic).
4. **`test_upc_asymmetric_pA`** — `lpp=2/2`, beam1 proton, beam2 Pb208,
   `pdlabel=edf1n2n`: must run (proton tag ignored) and give the same σ as `edfin2n`.
5. **`test_upc_o16_rejects_3n`** — O16 beams with `edf3n3n`: `InvalidRunCard` at
   card check, before any compilation.
6. **`test_upc_chff_o16`** — O16/O16 with `chf1n1n` runs (covers the ChFF branch and
   the O16 tables).

`tests/acceptance_tests/test_cmd_amcatnlo.py`:

7. **`test_upc_nlo_pbpb_edff`** and **`test_upc_nlo_pbpb_edpf`** — `a a > mu+ mu- [QED]`
   or the simplest available NLO UPC process, `launch -f` at fixed order with
   `pdlabel=edff` then `edpf`. The second one currently fails at run time because of
   the NLO makefile omission (B1) — it is the direct regression test for that bug.
8. **`test_upc_nlo_neutron_tag`** — `pdlabel=chf0nxn`, checks the run completes and
   σ < the untagged one.

`tests/acceptance_tests/test_cmd_reweight.py` / systematics:

9. **`test_systematics_upc_refused`** — `run systematics` on an `edf1n1n` LHE file
   exits with the "Systematics not supported" error rather than a traceback.

### C3. Minimal set if time is short

If only a handful can be added before merge: unit tests C1.1, C1.4, C1.9 and
acceptance tests C2.1, C2.2, C2.7. Those six cover the label set, the nucleus matrix,
the makefile linking (the found bug), the numerical baseline, the physics sum rule,
and the NLO path.

## B. Correctness / consistency findings

1. **NLO makefile misses `edpf/edmf/chpf/chmf`** (see A3). Real bug.
2. **U238 supported in Fortran, rejected by Python.** `photoabsorption.f90` implements
   Pb208, Au197, U238, O16 (the PR description also lists U-238), but the run-card
   check (banner + madevent_interface) only whitelists Pb, Au, O and raises for U.
   Either add `(92,146)` to the allowed nuclei or say in the PR why U238 is excluded.
   The run_card.dat comment says "only Pb208, Au197, O16".
3. **No Python-side nucleus check for NLO.** `PDLabelBlock` is in `RunCardLO.blocks`
   only; `RunCardNLO` has no equivalent. An NLO card with `pdlabel=edf3n3n` and
   `nb_proton=8` will only fail deep inside Fortran with `STOP` after compilation.
   A2's helper could be called from `RunCardNLO.check_validity` (using `pdlabel`).
4. **lpp sign.** Python checks `lpp1 == 2 == lpp2`; Fortran uses `ABS(LPP)`. `lpp=-2`
   is an allowed value in the LO run card. Either the check should use `abs()`, or
   -2 should be rejected for UPC. Pre-existing ambiguity, but the new check widens it.
5. **NLO run_card.dat comment.** The added block ends with
   `estimate none=No PDF, same as lhapdf with lppx=0` — the LO "none=..." line was
   glued onto the NLO text. `none` is not an allowed NLO pdlabel; drop that fragment.
6. **`get_pdf_id` comment** (banner.py 4004) claims the `except` is "to make sure
   edf<y>n<z>n ... are zero"; it is a generic fallback. Reword or list the UPC labels
   explicitly with the others (`'edff':0,'chff':0`).
7. **Dead branches** listed in A2.
8. **`raise Exception`** in madevent_interface should be `InvalidRunCard` (or removed).
9. **Version string / UpdateNotes.** `UpdateNotes.txt` is not touched; 3.8.1 notes
   should mention gamma-UPC 1.8 and the new labels.
10. `tests/input_files/run_card_quarkonium.dat` and `run_card_leptonium.dat` mention
    edff but are not referenced by any test (pre-existing orphans, noted in passing).

The gamma-UPC f90 core itself (interpolation-window rewrite, `NEUTRON_XSIGMA`,
new cross-section tables) is upstream code from the gamma-UPC authors; I did not
attempt to re-derive the numerics. The MG5 glue around it is what is reviewed above.

